PR #5, opened 2018-08-06
Add sort fn

alyssadelaine:


api/routes/users.js
```diff
@@ -59,6 +60,21 @@ module.exports = async (req, res) => {
       query = query.whereBetween('last_edit', [subMonths(Date.now(), 6), Date.now()])
     }
 
+    if (sortType.length > 0) {
+      if (sortType === 'Most recent') {
+        query = query.orderBy('last_edit', 'desc')
```
scisco: `query = query.orderBy('last_edit', 'asc')` is repeated multiple times. I wonder if there is a way to simplify this a bit.

CHANGELOG.md
```diff
@@ -5,10 +5,18 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+[Unreleased]: https://github.com/developmentseed/scoreboard/compare/v0.2.0...HEAD
+
+## [v0.2.0] - 2018-08-06
```
scisco: let's add these under unreleased for now.

PR #8, opened 2018-08-07
Api tests

alyssadelaine:


api/tests/test.js
```diff
@@ -0,0 +1,94 @@
...
+  nock(OSMESA_API)
+    .get('/users/1')
+    .reply(200, osmesa)
+})
+
+test.after.always(() => {
+  fs.removeSync(tempPath)
+})
+
+test('Test of OSMESA api call', async (t) => {
```
scisco: I think a more proper name for this test is something like `getting information of a single user` or something similar

api/tests/test.js
```diff
@@ -0,0 +1,94 @@
...
+  const res = await request(app)
+    .get('/scoreboard/api/users/1')
+    .expect(200)
+  t.is(res.body.id, '1')
```
scisco: We should test more features of the output. This ensures that the test fails if the output of the api is changed as a result of code changes

api/tests/test.js
```diff
@@ -0,0 +1,94 @@
...
+})
+
+test('Pull all users', async (t) => {
+  const response = await request(app)
+    .get('/scoreboard/api/users')
+    .expect(200)
+  t.is(4, response.body.records.length)
```
scisco: same as previous commment

api/tests/test.js
```diff
@@ -0,0 +1,94 @@
...
+
+test.before(async () => {
+  tempPath = path.join(os.tmpdir(), 'scoreboard', `${Date.now()}`, path.sep)
+  fs.mkdirpSync(tempPath)
+  process.env.TEST_DB_FILENAME = path.join(tempPath, 'db.sqlite3')
+
+  db = connection()
+  await db.migrate.latest()
+  // add a few records
+  await db('users').insert([{
```
scisco: I'd move the test users data to a json file and import here instead

api/tests/test.js
```diff
@@ -0,0 +1,94 @@
...
+const os = require('os')
+const path = require('path')
+const fs = require('fs-extra')
+const nock = require('nock')
+const test = require('ava') //eslint-disable-line import/no-unresolved
+const request = require('supertest')
+const connection = require('../db/connection')
+const app = require('../index')
+const osmesa = require('./fixtures/user_json.json')
+const mostRecentlyActive = require('./fixtures/most_recently_active')
```
scisco: it seems that most recently and least recently active users contain the same list. I recommend storing the user list as just a json file and the programmatically sort them in each test for validation.

Also please include `.json` in the require. It helps the reader realizes that this is a json file rather than a js module.

api/tests/fixtures/user_json.json
```diff
@@ -0,0 +1,88 @@
+{
+    "uid": 1,
```
scisco: I'd rename this file to `osma_api_user_output` to show that this is an output from an external system

api/tests/test.js
```diff
@@ -0,0 +1,94 @@
+'use strict'
+
+const os = require('os')
+const path = require('path')
+const fs = require('fs-extra')
+const nock = require('nock')
+const test = require('ava') //eslint-disable-line import/no-unresolved
```
scisco: the eslint-disable-line is not necessary. ava is added as dependency and should be found by the linter

api/tests/test.js
```diff
@@ -0,0 +1,94 @@
...
+  {
+    id: 2, osm_id: 3, edit_count: 5, display_name: 'B', country: 'Norway', last_edit: '250'
+  },
+  {
+    id: 3, osm_id: 4, edit_count: 10, display_name: 'C', country: 'Portugal', last_edit: '5'
+  },
+  {
+    id: 4, osm_id: 1, edit_count: 25, display_name: 'D', country: 'US', last_edit: '700'
+  }]
+  ) //eslint-disable-line function-paren-newline
```
scisco: the eslint disable would not be needed if the data is moved to a json file per previous comment

api/tm_clock.js
```diff
@@ -1,10 +1,12 @@
 const { TM_URL, TM_HASHTAG } = require('./config')
 const rp = require('request-promise-native')
-const conn = require('./db/connection')
 const {
   test, find, pick, merge, match, omit, tail
 } = require('ramda')
 
+const conn = require('./db/connection')
+const db = conn()
```
scisco: let's move this to the inside of the function (same as my previous comments)

api/users_clock.js
```diff
@@ -1,10 +1,11 @@
 const { USERS_URL, OSMESA_API } = require('./config')
 const rp = require('request-promise-native')
-const conn = require('./db/connection')
 const { compareDesc, parse } = require('date-fns')
 const {
   uniqBy, tail, zipObj, merge, map, props, sum, head, prop
 } = require('ramda')
+const conn = require('./db/connection')
+const db = conn()
```
scisco: same as my previous comment

alyssadelaine:
To Do list from @scisco's review:

- [ ] Put database connection inside try/function call

- campaigns.js
- topstats.js
- tm_clock.js
- users_clock.js

- [ ] Make sure json files have .json extensions in imports

- [ ] Rename user_json

- [ ] Rename "Test of OSMESA API call"

- [ ] Import mock user data from json

- [ ] Programmatically sort user data

api/tests/test.js
alyssadelaine: @scisco I wanted to make explicit the difference between calling the info in the database vs. calling the info from the api; would `get one user from API` work?

alyssadelaine: Decided to leave this as the original title per our discussion

alyssadelaine:
@scisco I think I addressed all your concerns, but could you take a look and let me know what you think?

PR #13, opened 2018-08-10
Feature/mock servers

kamicut:
Mocks some of the remaining servers so we don't need an env for development. In the future I want to join the fixtures with the test fixtures  / generate data instead of hard coding it.


scisco:
@kamicut this looks great, but please add a entry to the changelog.md

PR #17, opened 2018-08-22
Feature/users

kamicut:
This PR:
- switches the DB, tests, migrations, to postgres
- Allows logging in with OSM
- Adds some new tests fixtures
- Adds a new schema for users to remove things we can get from the OSM profile

This requires new environment variables for logging in with OSM and the local postgres DB


kamicut:
I'll have to fix the build so that it works on CI, but could you take a quick look @alyssadelaine @scisco for approach?

frontend/src/components/charts/HomeMap.js
```diff
@@ -15,14 +15,15 @@ class CampaignMap extends Component {
       renderWorldCopies: false,
       interactive: false
     });
+    console.log(this.props.overlay)
```
scisco: could you remove this?

frontend/src/App.js
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {Component} from 'react';
```
scisco: This has linting issues

kamicut: Where did you see the linting error?

alyssadelaine:
@kamicut I think "yarn clocks" should create the database if it hasn't yet been created; right now trying to run the commands in the order in the README gives an error

PR #20, opened 2018-08-30
add docs on how to run scoreboard locally

scisco:


kamicut:
Hey @scisco do you think we should have instructions on running a local postgres without Docker? I think it would be simple for people that already have it installed. Could you also add an entry in the changelog?

scisco:
@kamicut I think for consistency we should just use the docker postgres. This PR didn't have any changes to the code the require a changelog update.

kamicut:
I don't see how this is a consistency issue, since we don't require Docker before this PR and we should minimize dependencies for developers trying to contribute. I'll accept your changes and we can open this for another discussion.

PR #28, opened 2018-09-20
allow users to update data, starting with country

sethvincent:
There's a couple main todo items before merging this:

- make sure routes are protected so only the authenticated user can update their own profile
- add a link to the user edit page

### Demonstration of user edit page:

scoreboard-edit-profile

### DIFF
--- a/api/src/db/seeds/utils.js
+++ b/api/src/db/seeds/utils.js
@@ -8,7 +8,6 @@ function generateUsers(n, knex) {
   const users = []
   for (let i = 0; i < n; i += 1) {
     const user = {
-      id: i + 1,
       osm_id: 100000000 + i,
       full_name: `test${i} McTest`,
       country: countryIds[parseInt(Math.random() * (countryIds.length - 1))]
--- a/api/src/models/users.js
+++ b/api/src/models/users.js
@@ -6,11 +6,11 @@ class Users {
   }
 
   get(id) {
-    return this.db('users').where('id', id).then((rows) => rows[0])
+    return this.db('users').where('id', id)
   }
 
   findByOsmId(osmID) {
-    return this.db('users').where('osm_id', osmID).then((rows) => rows[0])
+    return this.db('users').where('osm_id', osmID)
   }
 
   find(key, value) {
--- a/api/src/models/users.js
+++ b/api/src/models/users.js
@@ -1,7 +1,7 @@
 const connection = require('../db/connection')
 
 class Users {
-  constructor(options) {
+  constructor() {
     this.db = connection()
   }
 
--- a/api/src/models/users.js
+++ b/api/src/models/users.js
@@ -1,0 +1,33 @@
+const connection = require('../db/connection')
+
+class Users {
+  constructor(options) {
+    this.db = connection()
+  }
+
+  get(id) {
+    return this.db('users').where('id', id).then((rows) => rows[0])
+  }
+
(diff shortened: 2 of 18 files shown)


api/src/db/seeds/utils.js
```diff
@@ -8,7 +8,6 @@ function generateUsers(n, knex) {
   const users = []
   for (let i = 0; i < n; i += 1) {
     const user = {
-      id: i + 1,
```
sethvincent: I had to do this because otherwise any new users were starting at 1.

api/src/models/users.js
```diff
@@ -0,0 +1,33 @@
+const connection = require('../db/connection')
+
+class Users {
```
sethvincent: So far this model is only used to create new authenticated users, but we could reuse it in a few places.

api/src/passport.js
```diff
@@ -30,18 +32,39 @@ if (NODE_ENV === 'test') {
...
+  })
 
   passport.use(new OSMStrategy({
     consumerKey: OSM_CONSUMER_KEY,
     consumerSecret: OSM_CONSUMER_SECRET,
     callbackUrl: `${API_URL}/auth/openstreetmap/callback`
-  }, (token, tokenSecret, profile, done) => {
-    if (profile) {
-      return done(null, profile)
+  }, async (token, tokenSecret, profile, done) => {
```
sethvincent: Should we be storing the token/tokenSecret? Would we be able to login using those so that we don't have to grant oauth access on every login? I'll look into the openstreetmap provider to try to better understand how that works.

frontend/src/App.js
```diff
@@ -114,6 +126,7 @@ class App extends Component {
           <Route exact path="/home" component={Home} />
           <Route exact path="/campaigns" component={Campaigns} />
           <Route exact path="/users" component={Users} />
+          <Route exact path="/edit/:uid" component={UserEdit} />
```
sethvincent: I tried to make this route `/users/:uid/edit` or `/users/edit/:uid`, but because of how the `/users/:uid` container works neither of those were possible.

api/src/passport.js
kamicut: I think we can, so that we don't need to grant permissions to OSM every time. Right now if you logout and log back in, then you have to "grant permissions" again.

frontend/src/containers/Dashboard.js
```diff
@@ -50,6 +50,7 @@ class Dashboard extends Component {
     }
 
     // We are logged in and should have a profile
+    console.log('profile', profile)
```
kamicut: We should remove this console.log

frontend/src/App.js
```diff
@@ -94,6 +105,7 @@ class App extends Component {
                             <ul>
                               <li><ActiveLink to="/dashboard" label="Dashboard" /></li>
                               <li><ActiveLink to={`/users/${profile.id}`} label="Public Profile" /></li>
+                              <li><ActiveLink to={`/edit/${profile.id}`} label="Edit Profile" /></li>
```
kamicut: Note: We should probably rethink the router here and make this:
1. An authenticated only route
2. Part of the dashboard page

frontend/src/App.js
kamicut: Fine for now, but once this is merged let's open a new PR for how we want to merge this functionality into the /dashboard page.

PR #30, opened 2018-09-24
Pull badges from the DB and calculate for each user on the backend

alyssadelaine:
Pulling from the database is still in progress


alyssadelaine:
Actually @kamicut I think I should do some more things before we merge this

frontend/src/components/BadgeBox.js
```diff
@@ -51,12 +27,14 @@ function stripWS(text) {
 export default ({ badges }) => {
   if (badges && badges.all) {
     var progressBadges = Object.keys(badges.all).map(val => {
+      console.log(badges)
```
kamicut: remove these console.log

api/src/badge_logic/badges.js
```diff
@@ -1,51 +1,52 @@
-export default {
+module.exports = {
```
kamicut: Why do we still need this file?

alyssadelaine: You're right, we don't. I'll remove this and the console.log @kamicut

PR #34, opened 2018-09-25
Use <uid> as source-layer

mojodna:
An upcoming change to user footprint generation in OSMesa changes MVT layer names to match their corresponding key (user id or hashtag, depending on context). This is being done to facilitate merging multiple users' footprints together into a single MVT (and to limit the number of tiles needing to be managed).

This should be deployed in conjunction with a configuration change to https://github.com/azavea/osmesa-stat-server (facilitated by @moradology) that will point footprints at their latest revision.


mojodna:
Hold in merging this. We're switching usernames to user IDs, as they don't have the potential to change.

mojodna:
Ok, this is ready to merge again (but not to deploy, pending correcting the tiles and deploying the configuration change).

scisco:
this was included and merged as part of #42

PR #43, opened 2018-10-09
Bring over v0.2 patches to unreleased version

scisco:


frontend/src/utils/api.js
```diff
@@ -3,8 +3,8 @@ var debounce = require('debounce-promise');
 const {isNil} = require('ramda');
 
 const instance = axios.create({
-  baseURL: '/scoreboard'
-})
+  baseURL: window.location.pathname
```
sethvincent: Will this stop working if we switch from using the hash router to the pathname router? It seems like it.

I'm not sure when that switch will happen, though.

PR #49, opened 2018-10-16
user roles & admin functionality

sethvincent:
This pr adds user roles to the project.

It also sets up authentication mocking so we can run tests that involve logged in users. I used [passport-mock-strategy](https://github.com/cszatma/passport-mock-strategy) for this. We may need to add to the mock user that is created depending on the other testing needs that come up.

todo before this is ready:
- [x] fix running tests on CI
- [x] lint fixes
- [x] fix api mistakes
- [x] update changelog
- [ ] some ui implementation


api/test.sh
```diff
@@ -1,7 +1,9 @@
 if [ -z "$CI" ]; then
-    createdb -p 5433 scoreboard_tests && 
-    ava tests/*.test.js ; 
-    dropdb  -p 5433 scoreboard_tests
+    docker-compose exec db createdb -U postgres scoreboard_tests
```
sethvincent: I had a hard time getting tests to run and made these changes.

Not sure if there are any issues here, but this is working well locally.

I need to revise the `else` part of this statement to work on CI.

api/bin/commands/role/set.js
```diff
@@ -0,0 +1,61 @@
...
+    return process.exit(1)
+  }
+
+  if (!role) {
+    console.error('role is required')
+    return process.exit(1)
+  }
+
+  const users = new Users()
+  console.log('users', users)
```
kamicut: Remove this console.log?

api/src/models/users.js
```diff
@@ -1,4 +1,5 @@
 const connection = require('../db/connection')
+const { validateRole } = require('../utils/roles')
 
 class Users {
```
kamicut: To add a role, do we just call `update`? Do we want to make it more explicit than that?

kamicut:
@sethvincent this PR is affected by the merge @alyssadelaine has made in another branch. Let me know if you need help there. Also, do you want to move UI additions to another PR? I am wondering if we need a UI so that an admin sets roles for other users

PR #52, opened 2018-10-19
fix admin user editing

sethvincent:
I noticed a bug with user editing where the edit form had the wrong id. This PR fixes that.


kamicut:
@sethvincent is this something that can be covered by a test?

PR #54, opened 2018-10-22
allow setting custom project names

scisco:


kamicut:
@scisco what does unregistering the service workers do?

scisco:
@kamicut if there service worker is registered on a browser for a website, there is no easy way to remove it unless you include code to unregister it (for example hard refresh and clearing cache does not remove it).

This causes problem for anyone who has used a previous version of scoreboard. The previously deployed service worker was messing with the user endpoint on the api and caused the dashboard to show 0s for the total number of users. 

The unregistering fixes that.

PR #59, opened 2018-10-23
Add back in obtainable badge list. Should fix dashboard error

alyssadelaine:


kamicut:
This does indeed fix the dashboard error, but I'm getting something else:



I'm wondering, should we not have most attainable badges and just render all the badges that you didn't get? cc @sethvincent

sethvincent:
The `NaN` is happening because the badge objects used to have properties at `points.nextPoints` and `points.currentPoints`. There's a `points.percentage` property and it is at 100% for the badges that are showing up. I'm seeing that the notion of `tiers` is not in the schema now so that makes sense.

We could show all badges that haven't been earned yet, but it would still be useful to sort by which are closest to being done.

Here's what I'm thinking:

- remove mostAttainable again 🙃
- filter sortedBadges to exlude any that are at 100%
- return that as unearnedBadges
- take the top three and show them in the dashboard (and later we can surface the others as well)

sethvincent:
screen shot 2018-10-24 at 11 33 06 am

PR #62, opened 2018-10-24
[wip] feature/next

kamicut:
@sethvincent @scisco @alyssadelaine here are my initial thoughts on the restructuring to work with `next`. This is a wip PR because it touches a lot of files.

Backlog:
- [x] Install next scaffolding
- [x] Get simple page to render
- [x] Get API docs to render
- [x] Try a page with data from the API
- [x] Make redux / unistore work 
- [x] Make authentication work
- [x] Render all pages
- [x] Make tests work
- [x] Write documentation
- [ ] Figure out how to merge to master 😅


kamicut:
Woops merged in a .env, going to revoke those creds

kamicut:
The homepage now renders

sethvincent:
Most tests are now running locally. I fixed a bug in the roles migration that wasn't coming up before.

There's a very confusing bug happening in the roles test where `/auth/openstreetmap` returns a 404 when creating an admin user.

Something about the mock passport strategy isn't working, though I'm not sure what. Requests to other routes like `/auth/userinfo` work as expected.

For now I've commented out the api tests in api/tests/roles.test.js.

So far I've ran the tests against a postgres db in docker, run with docker-compose.

I've added scripts to create a second db in docker that tests run against. It should also work when using `DATABASE_URL` to point to a local postgres db, though I haven't tested that yet.

PR #64, opened 2018-11-07
Support for date-specific and campaign-specific badges with task descriptions

alyssadelaine:
Added a 'between' operator which requires a longer list in the database. This was the only way to support date ranges with a begin and end time

Task description creation is also slightly different for date badges because they introduce the idea of the expiration of badge eligibility, which is a new concept to scoreboard

Rewrote logic variables in get_badge_info to be more transparent


alyssadelaine:
screen shot 2018-11-07 at 9 04 21 am

alyssadelaine:
This badge is an example where a range was set from 2018-11-06 to 2018-11-08
screen shot 2018-11-07 at 11 25 20 am

sethvincent:
@alyssadelaine I changed the base for this pr to master since the feature/next branch is now merged.

alyssadelaine:
Thanks @sethvincent ! Do I need to do something to make the tests pass?

sethvincent:
@alyssadelaine Once the conflicts are resolved I expect the CI will pass.

PR #96, opened 2018-11-28
Get all pages of results from TM 3 apis

alyssadelaine:


api/src/tm_clock.js
```diff
@@ -52,41 +52,49 @@ async function tmWorker () {
...
-              geometry,
-              name: feature.name,
-              description: feature.shortDescription,
-              validated: feature.percentValidated,
-              done: feature.percentMapped,
-              tm_id: feature.projectId
-            }
-          })
+        let sqlPromises = []
+        for (let page = 1; page <= json.pagination.pages; page++) {
```
scisco: Using for loop with await means that we are making requests in sequence and not taking advantage of the async natures of js.

I suggest we use promise.all with along with a library like `p-limit` to control the concurrency. This should also considerably speed up the the process.

kamicut:
Work merged into #217

PR #97, opened 2018-11-28
Randomize last_edit

alyssadelaine:


scisco:
@alyssadelaine this change seems to be breaking the tests. Could you fix?

alyssadelaine:
@scisco Yeah I saw that; I think it's because the databases are different, so randomizing the seeds returns different values in the tests. Is it a priority to fix this if it just applies to test data?

alyssadelaine:
Closing in favor of https://github.com/developmentseed/scoreboard/pull/151

PR #115, opened 2018-11-29
fix development knex config

sethvincent:
`def` was overriding the `development` config so `production` seeds were being run in development.

### DIFF
--- a/api/src/db/knexfile.js
+++ b/api/src/db/knexfile.js
@@ -37,6 +37,5 @@ module.exports = {
     },
     debug: process.env.KNEX_DEBUG || false
   }),
-  staging: def,
   production: def
 }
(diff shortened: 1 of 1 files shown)


api/src/db/knexfile.js
```diff
@@ -31,12 +31,12 @@ const def = {
...
   test: test,
-  development: Object.assign({}, {
+  development: Object.assign(def, {
     seeds: {
       directory: path.join(__dirname, 'seeds', 'development')
     },
     debug: process.env.KNEX_DEBUG || false
-  }, def),
+  }),
   staging: def,
```
kamicut: @sethvincent @scisco we don't need a staging def anymore, right? We can remove it in this PR as well. 
```suggestion
```

Issue #131, opened 2018-12-04
Filters, sorts, and search on Users Table are not working


alyssadelaine:
![screen shot 2018-12-04 at 10 24 39 am](https://user-images.githubusercontent.com/6155420/49452314-f9ac7800-f7ae-11e8-82df-073c93905204.png)

PR #135, opened 2018-12-04
Fix/users

kamicut:
@sethvincent @alyssadelaine this fixes the users page filtering / sorting


alyssadelaine:
Fixes https://github.com/developmentseed/scoreboard/issues/131

PR #138, opened 2018-12-05
Feature/admin teams

kamicut:
Admins can create / update / delete teams interacting with the OSM Teams service. @scisco I could use some help figuring out a minimal configuration to deploy Scoreboard since OSM Teams is not public yet.


scisco:
@kamicut let's do a quick voice on that and I'll help

scisco:
osm-teams is now public 🎉 

cc @sethvincent

PR #148, opened 2018-12-07
Feature/teams page

kamicut:
This adds a teams list, an individual team page and the ability for an admin to assign campaigns to a team

### DIFF
--- a/api/src/db/migrations/20181207115544_team_assignments.js
+++ b/api/src/db/migrations/20181207115544_team_assignments.js
@@ -1,0 +1,21 @@
+
+exports.up = async function (knex) {
+  try {
+    await knex.schema.createTable('team_assignments', table => {
+      table.increments('id')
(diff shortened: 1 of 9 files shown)


sethvincent:
An edge case but it would be nice to just say "no teams yet" when there's no teams on the teams page

screen shot 2018-12-07 at 1 32 29 pm

sethvincent:
Campaigns should be removed from the unassigned list when they are assigned to a team. There should also be a little less white space there (sorry I should have seen this in the teams admin pr).

screen shot 2018-12-07 at 1 46 22 pm

kamicut:
@sethvincent that's strange, i'm getting "0 results" when there are no teams

sethvincent:
Ah yeah it happened when I had osm-teams in a bad state. I don't think it showed an error message.

PR #154, opened 2018-12-10
Limit participants to top 15 on country pages

alyssadelaine:
screen shot 2018-12-10 at 5 54 33 pm


alyssadelaine:
@kamicut Yeah I thought about that but I wasn't sure if we would need the full list at some other point
I'll do it with an optional argument

Issue #133, opened 2018-12-04
Badge management lacks field for uploading badge icon


scisco:
There is no field for adding or updating an icon associated with a badge

alyssadelaine:
The badge images we currently have are associated with metric categories (for example, "Waterways"), so we could use that to allow admins to choose which image they want to be displayed with their custom badge

kamicut:
@alyssadelaine what are you thinking the component could look like for the badge icon selection?

PR #169, opened 2018-12-11
Updates to badge admin pages

alyssadelaine:
Closes https://github.com/developmentseed/scoreboard/issues/133

* Adds a carousel with available badge images for admin to choose from at time of badge creation
* The same carousel (scrolled to the previous image) is available to choose a new image when editing a badge
* Allows an admin to enter the name of a hashtag and create a badge for users who have edited using that hashtag
* Changes the form for hashtag badges since "Condition" and "Number" aren't meaningful in that case


lib/utils/loadImages.js
```diff
@@ -0,0 +1,18 @@
+function createImageList () {
```
kamicut: This is pretty opaque, what are you doing here? Could you add a comment?

alyssadelaine: done

kamicut:
@scisco @sethvincent could you try this in your browsers? @alyssadelaine and I can get it working on Chrome and some version of Firefox, but I couldn't get it running in my current Firefox.

sethvincent:
Worked for me in firefox, chrome, and safari.

We should move the "add another condition" button to be above the image selection. It's a little confusing being separated from the existing conditions.

Issue #173, opened 2018-12-12
Add more varied seed user data


kamicut:
Right now seed user data is generated from the same sample user. We should have randomized user (including 0s in some fields) to get a better sense of how different components look with "real-looking" data.

PR #178, opened 2018-12-12
Fix source of truth for country names across app

### LINKED ISSUE #167: Use country full name instead of alpha codes
* Using alpha codes is unsustainable with OSM
* Also need to change users_clock function to avoid double-counting country edits

kamicut:
- Closes #167 by moving to a single source of truth with natural earth countries.geojson from https://github.com/azavea/osmesa/blob/master/src/analytics/src/main/resources/countries.geojson

- Takes a stab at #173 by generating new user data


api/src/models/userCountryEdits.js
```diff
@@ -1,9 +1,4 @@
 const db = require('../db/connection')
-const util = require('util')
```
scisco: This is needed in the clocks file for users. It either has to be added to user clocks or be restored here.

api/src/users_clock.js
```diff
@@ -43,9 +43,6 @@ async function updateCountries (userID, countryEditList) {
   // get total edits for a given user
   countryEditList.forEach((tuple) => {
     let countryName = tuple.name
-    if (isState(countryName)) {
```
scisco: This is essential for marking all US states under the US name

PR #195, opened 2018-12-17
Close #189

### LINKED ISSUE #189: Badges should be forced to have unique names
Currently I can create two "Consistency" badges, which is confusing and not the desired behavior

kamicut:
This fixes the uniqueness constraint at the database, route and store level. I propagate the error down to the store and then catch the error before we submit the form. We could replicate this pattern for other form submits, and ideally we would have a HOC for this. 

I also added tests for the constraint and fixed something about the POST request (it wasn't sending the id of the newly created badge)


pages/admin/badges-add.js
```diff
@@ -479,4 +481,4 @@ export class AdminBadgesAdd extends Component {
   }
 }
 
-export default connect(['authenticatedUser', 'error', 'badges', 'admin'], actions)(AdminBadgesAdd)
+export default connect(['authenticatedUser', 'error', 'badges', 'admin', 'notification'], actions)(AdminBadgesAdd)
```
sethvincent:
```suggestion
export default connect(['authenticatedUser', 'error', 'badges', 'admin'], actions)(AdminBadgesAdd)
```

The 'notification' property of state isn't needed.

pages/admin/badges-edit.js
```diff
@@ -502,7 +503,7 @@ export class AdminBadgesEdit extends Component {
   }
 }
 
-const Page = connect(['authenticatedUser', 'error', 'badge', 'admin'], actions)(AdminBadgesEdit)
+const Page = connect(['authenticatedUser', 'error', 'badge', 'admin', 'notification'], actions)(AdminBadgesEdit)
```
sethvincent:
```suggestion
const Page = connect(['authenticatedUser', 'error', 'badge', 'admin'], actions)(AdminBadgesEdit)
```

The 'notification' property of state isn't needed.

Issue #185, opened 2018-12-13
Frontend user page URL should include username rather an ID


alyssadelaine:
So instead of /users/123456, the URL should be /users/TheirUserName

kamicut:
This one might be tricky, OSM usernames can change (and OSM ids don't), so we have to make sure the latest username in the OSMesa data is associated to the proper OSM id. Similar issue in missingmaps https://github.com/MissingMaps/missingmaps.github.io/issues/265

PR #202, opened 2018-12-18
Close #197

### LINKED ISSUE #197: Link to user page from team page
Add hyperlink to a user's public page from the `Team Members` list - currently no stats on the Team Page, so we want to be able to see how a user is contributing

kamicut:
Adds a link to the users on the teams page


alyssadelaine:
Note that this implementation might need to change based on https://github.com/developmentseed/scoreboard/issues/185

Issue #201, opened 2018-12-18
Tailor some features of the dashboard based on user role


alyssadelaine:
For example, as an admin I would like to just click a link to create/view badges straight from the dashboard

alyssadelaine:
Closed in ^

PR #214, opened 2018-12-20
add admin section list to dashboard if user is admin

sethvincent:
Links to admin pages now appear in a component on the dashboard if the user is an admin. I've reused the same component on the admin index page.

this might be all we need for #201 but open to other improvements as well.

screen shot 2018-12-20 at 10 44 18 am

### DIFF
--- a/components/AdminSectionList.js
+++ b/components/AdminSectionList.js
@@ -1,0 +1,55 @@
+import React from 'react'
+import Link from './Link'
+
+export default () => {
+  return (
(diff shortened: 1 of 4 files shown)


alyssadelaine:
@sethvincent I think instead of "[Feature] list", those links should be "See all [feature]s" -- it's a little more exciting and clear, and it keeps with the active tense of "Create a new badge"

So the first row would be:
See all badges
See all teams
See all users

PR #217, opened 2019-01-02
Feature/admin tasks

kamicut:
- [x] REST API to add different tasking managers
- [x] Move TM service code to abstract base class
- [x] Move clock code to use TM services
 - [x] Create new overloadable function `runTMUpdate` in each class
 - [x] Create async task runner that wraps `runTMUpdate` for array of TMs from DB
- [x] Fix extract campaign hashtag
- [x] Tests for TM2 and TM3
- [x] Add ability to delete tasking manager with all its campaigns


kamicut:
This is tested for TM2, I still need to test for TM3

pages/admin/taskers-add.js
```diff
@@ -0,0 +1,268 @@
...
+    return (
+      <div className='admin'>
+        <AdminHeader />
+        <section>
+          <div className='row'>
+            <div className='sidebar'>
+              <h2 className='header--large'>Tasking Managers</h2>
+              <ul className='admin-sidebar-links'>
+                <li>
+                  <Link href='/admin/badges'>
```
sethvincent: this should go to the tasking manager list

pages/admin/taskers.js
```diff
@@ -0,0 +1,108 @@
...
+        <AdminHeader />
+        <section>
+          <div className='row'>
+            <div className='sidebar'>
+              <h2 className='header--large'>Tasking Managers</h2>
+              <ul className='admin-sidebar-links'>
+                <li>
+                  <Link href='/admin/taskers/add'>
+                    <a className='link--large'>
+                    Add new Source for Tasks
```
sethvincent: I'd expect this to be

```suggestion
                    Add new tasking manager
```

sethvincent:
There are a few places where `tasker` is used as shorthand. I'd instead use the full `tasking-managers` in the url and `tasking manager` in the save button. It's longer but is more clear/consistent.

kamicut:
@alyssadelaine @sethvincent for TM3 geometries I'm a bit conflicted; in TM2 to get the boundaries of tasks you could get it as one request per _page_ which was is fine (if you have 300 campaigns, that's 3 requests). In TM3 you would need to do 300 additional requests.

TM3 gives us the map centers of all the tasks in 1 request, which could be enough for the cards, but not the boundaries. I think if we can get the center for the cards, but then get the boundaries when on each project page that would be less requests to the tasking manager.

components/AdminSectionList.js
```diff
@@ -50,6 +50,18 @@ export default () => {
...
+      <li>
+        <h3 className='header--medium'>
+          Tasking Managers
+        </h3>
+        <ul>
+          <li>
+            <Link href='/admin/tasking-managers'>
+              <a className='link--normal'>See all tasking managers</a>
+            </Link>
+          </li>
```
sethvincent: We should add a link here for adding a new tasking manager

sethvincent:
Tested locally and things are working well with a couple small issues.

Steps I took (writing down here because we could document these kinds of steps more often):

- reset the db `npm run rollback && npm run migrate && npm run seed`
- reinstalled modules `npm i`
- started the server `npm run dev`
- added myself as admin `scoreboard create-admin-user -u sethvincent`
- added tasking managers via admin ui
  - https://tasks.hotosm.org/
  - https://tasks.hotosm.org/
- went to campaigns page and wondered why there were none
- realized i should run `npm run clocks` so did that
- explored around the campaigns

### Urgent issue:

- when I click `contribute` from a campaign page for a TM3 project I'm being sent to a url like: https://tasks.openstreetmap.us/project/1910
  - I think that button needs to be updated to reflect the source of the campaign

### A few questions/suggestions that may need to be future improvements:

- It would likely be useful to show on a campaign page which tasking manager was the source
- It would be cool to filter by tasking manager on the campaigns page.
- Could we add unit tests using something like https://github.com/flickr/yakbak to record the api responses or using sample data? It would be nice to be able to make sure the the tm services modules continue to work. Unit tests for the model/routes would also be good.
- Users will be surprised when they add a tasking manager, go to /campaigns, and not see any new campaigns. The clocks process will update eventually, but maybe we could kick it off manually? Ideally just for that one new tasking manager? At minimum we should have some kind of help text somewhere that says campaigns are imported on a schedule.

kamicut:
> Could we add unit tests using something like https://github.com/flickr/yakbak to record the api responses or using sample data? It would be nice to be able to make sure the the tm services modules continue to work. Unit tests for the model/routes would also be good.

I'm not sure I understand what I'd be testing for? Could you describe a test? Also would you want it as part of this PR?

> Users will be surprised when they add a tasking manager, go to /campaigns, and not see any new campaigns. The clocks process will update eventually, but maybe we could kick it off manually? Ideally just for that one new tasking manager? At minimum we should have some kind of help text somewhere that says campaigns are imported on a schedule.

Would adding a column for "Last ran on x date" for each tasking manager in the admin list be enough? I initially thought that it would be great to run the TM from the admin panel, but I realized that without a queuing or messaging server like Redis the communication between a subprocess and the server would be a bit kludgy: how would we know when the process ends, or if we should retry, etc? 

> when I click contribute from a campaign page for a TM3 project I'm being sent to a url like: https://tasks.openstreetmap.us/project/1910

The specific domain tasks.openstreetmap.us? Must be a hardcoded value somewhere

sethvincent:
### tm service testing
For testing the tm service types it could be pretty simple: given a set of campaigns that match what we expect for TM3 for example, make sure they can be imported to the db.

Just so that if we make changes later to tmWorker or the tm service types we can run a quick test to make sure nothing broke instead of having to set up tasking managers in the ui and manually run the clocks process, and then deleting the db and repeating that process.

### campaign import
Yeah something like `Campaigns imported ten minutes ago` or `Campaigns not imported yet` might be good, plus some help text on the add form that says something like `Creating a tasking manager will schedule an import of campaigns. Imports are run every 10 minutes.`

I agree, trying to kick off an import manually or on tasking manager creation is out of scope for this PR.

api/src/routes/campaign.js
```diff
@@ -13,19 +14,28 @@ const db = require('../db/connection')
  * @returns {Promise} a response
  */
 module.exports = async (req, res) => {
-  const { id } = req.params
-  if (!id) {
+  const { tasker_id, tm_id } = req.params
+  console.log(req.params)
```
sethvincent:
```suggestion
```

api/src/routes/campaign.js
```diff
@@ -13,19 +14,28 @@ const db = require('../db/connection')
...
+  if (!tm_id || !tasker_id) {
     return res.boom.badRequest('Invalid id')
   }
 
   try {
-    const osmesaResponse = await osmesa.getCampaign(id)
-    const [tmData] = await db('campaigns').where('campaign_hashtag', id)
+    const [tmData] = await db('campaigns').where({ tasker_id, tm_id })
+    console.log(tmData)
```
sethvincent:
```suggestion
```

pages/admin/tasking-managers-edit.js
```diff
@@ -0,0 +1,346 @@
...
+        nameInput: tasker.name,
+        descriptionInput: tasker.description,
+        urlInput: tasker.url,
+        typeInput: tasker.type
+      })
+    }
+  }
+
+  async updateTasker (params) {
+    console.log(params)
```
sethvincent:
```suggestion
```

alyssadelaine:
Can we provide some info to admins on when the next clocks cycle will be run, even if it's just "within 10 min"?
And shouldn't the default TM still show up on the admin panel?

alyssadelaine:
I added https://tasks.openstreetmap.us as a TM URL with TM 3 and I get "Failed to get Campaign Data" on every related campaign

kamicut:
> Can we provide some info to admins on when the next clocks cycle will be run, even if it's just "within 10 min"?
I don't see how we can do this without having to do interprocess communication with the clocks, which is out of scope of this PR

> Shouldn't the default TM still show up on the admin panel?
I don't understand this question


> Can we add filter functionality to the Campaigns page?
Sure, I can add that, @LanesGood could we work on how that would show up in the sidebar? 

I'll also look into why the URL doesn't work with a slash

alyssadelaine:
> I don't see how we can do this without having to do interprocess communication with the clocks, which is out of scope of this PR

Right I wasn't trying to imply that any heavy code lift should be done here; just that within the admin panel instead of "Never" we should put something like "To be updated within 10 minutes" since we plan to run the clocks process every 10 min

> > Shouldn't the default TM still show up on the admin panel?

> I don't understand this question

I misunderstood how this functionality was being integrated, so this question is moot

kamicut:
@alyssadelaine ok I'll add some more help text. About the slashes, I think we can sanitize the user input with a simple regex replace for trailing slashes. However, I just tested a tasking manager with a trailing slash to see what would happen and did not get any errors. I'm wondering if there's something else going on with your setup. The `null-80` refers to `:tm_id-:project_id` so why didn't your tm get added, or has an id of null 🤔 ?

alyssadelaine:
@kamicut I don't think so, because it worked later. But it's possible the ending slash was a red herring

alyssadelaine:
@kamicut "Last updated" shows as `Over 49 years ago` for all the campaigns
screen shot 2019-01-24 at 12 16 54 pm

Non-blocking comment: I added a tm meant to error and it still shows up as the admin as "Will run in the next 10 min". Would be nice if it updated to say there was an error with this tm which should be resolved

kamicut:
> Non-blocking comment: I added a tm meant to error and it still shows up as the admin as "Will run in the next 10 min". Would be nice if it updated to say there was an error with this tm which should be resolved

This is also something that could be solved by having  interprocess communication between the API and the clocks process. However right now the API doesn't know anything about the clock runs. We could catch the error in a new column for `taskers` called "STATUS".

kamicut:
@alyssadelaine the reason that updated_at was not working is there was no good way to bulk get lastUpdate for TM3. For now, I'm calling the TM for the project data when the page loads. This slows down the server response but we can at least get this number.

PR #221, opened 2019-01-09
add/edit badges with date-specific criteria

### LINKED ISSUE #218: manage date-specific badges in admin ui
Based on the [related back-end work](https://github.com/developmentseed/scoreboard/pull/64) the badges forms in the admin ui need to support creating and editing date-specific badges.

sethvincent:
Changes include:

- UI for adding & editing badges with date-specific requirements
- A few small changes on the server side to get things working
- Minimal refactoring of badge-add.js page so that we're not switching operations between objects and arrays which made things a little easier for working with the date operations and reduced differences between badge-add.js and badge-edit.js

Example of the input fields for date-specific metrics:

screen shot 2019-01-10 at 2 05 40 pm

So far have tested creating and updating badges with a bunch of different combinations via the UI.

Steps for testing included the usual:

- reset the db `npm run rollback && npm run migrate && npm run seed`
- reinstalled modules `npm i`
- started the server `npm run dev`
- added myself as admin `scoreboard create-admin-user -u sethvincent`
- added and edited badges via the admin ui
- checked out the api endpoint to verify the data is right

Closes #218


sethvincent:
So far this is implemented in the form for adding badges. Working on editing badges now.

alyssadelaine:
@sethvincent To keep the UX consistent, I think we should continue to use the condition field for the date-specific badges. So those options can read "before", "after", "on," or "between", and in the last case a second date field should be added below

api/src/badge_logic/get_badge_info.js
```diff
@@ -70,7 +70,7 @@ function simpleDateComp (requiredDay, days, operator, today, excluded) {
   const validDate = new Date(validDay[0], Number(validDay[1] - 1), validDay[2])
   if (!(execLogic(operator, today, validDate))) {
     return [ pass, true, '' ]
-  } else if (operator[0] === '<') {
```
sethvincent: I was having operators other than `<` hit this and because they were unhandled this function returned `undefined`.

alyssadelaine: The idea behind that was that the first condition checks to see whether mapping today would not contribute to the badge, and if it would contribute to the badge, then the "due by" date must be in the future.

I'm assuming the operator that hit the else was "=" when you were testing with validDate = today? I think it makes sense to provide for that, but if other operators got through then something else may be going on.

PR #224, opened 2019-01-10
Separate completeness and validation sliders

### LINKED ISSUE #211: Completeness slider should measure "Complete" parameter
@kamicut I remember us talking about how the Completeness slider measures a combination of "Complete" and "Validated", but it's confusing to set the "Completeness" range between 0 and 20 and see a 22% "Complete" campaign. I think we should add a slider for Validation

kamicut:
Close #211

Also changes color of sliders to match percentages

image


alyssadelaine:
@kamicut How difficult is it to remove the gray "0" and "100" when the slider endpoints are on those numbers? It looks strange to have the same number above and below the slider

kamicut:
image

@alyssadelaine the newest commit removes the gray legend

PR #231, opened 2019-01-16
allow setting docs base url using env var

sethvincent:
By setting the docs baseUrl to something like `/example/docs/` it will create a build where hrefs and other links use `/example/docs/` as part of the url.


alyssadelaine:
This PR fixes the between-pages links, but the links back to Scoreboard and the images/styles are still broken. @sethvincent , any idea why that would be the case?

kamicut:
Closing this for now until we can figure out how to merge app, api docs and user docs

PR #232, opened 2019-01-16
Add command to run users and clocks

alyssadelaine:


kamicut:
@alyssadelaine what is the situation where we need to run clock_updates vs clocks?

alyssadelaine:
👍 That's fine with me

PR #239, opened 2019-01-24
Replace WebGL with Leaflet

kamicut:
This PR replaces the WebGL maps with Leaflet maps using `react-leaflet`. Before merging, I think @LanesGood should look over the CSS i made up and stuck in index.scss. 

On the dashboard and user profile pages, I do a lookup in `countries.geojson` using the OSMesa `country_list` and construct a featurecollection for the user's edited countries. This could be slow if a user has edited in multiple countries. To render it I have a _very_ basic formula for opacity, and we should probably look at either creating opacity breaks or color breaks (with a legend?)

Closes #239


kamicut:
I haven't changed the underlying style, so this is a general style decision @LanesGood

PR #241, opened 2019-01-25
Fix/country listings

kamicut:
* Fixes the homepage country listing to display name of countries mapped (instead of edits by mapper country) as well as links to those countries
* Fixes the dashboard country listing in the sidebar


api/src/routes/topstats.js
```diff
@@ -49,10 +49,11 @@ module.exports = async (req, res) => {
     const [{ totalEdits }] = await db('users').sum('edit_count as totalEdits')
 
     const [{ numUsers }] = await db('users').count('id as numUsers')
-    const editsByCountry = await db('users')
-      .whereNotIn('osm_id', filteredUsers)
```
alyssadelaine: Shouldn't we still check filteredUsers @kamicut ?

kamicut: Oh, that's a good point, I'll amend the sql

PR #251, opened 2019-01-30
Exclusion list

### LINKED ISSUE #248: Interface to exclude mappers from results
Right now we exclude mappers from results using an environment variable. It would be better if this was handled in the Admin interface.

kamicut:
Closes #248 

This adds an admin interface to add / remove mappers from a blacklist (exclusion list). This would for example remove bots from the calculation of statistics in the users page and the homepage.

*TODO*
- [x] @LanesGood if you could take a look at my markup and see if anything is wrong there
- [ ] Documentation when design is done


LanesGood:
@kamicut I've modified the table styling slightly. 

I also suggest that the button for the "modify excluded users" form be moved just below the selected users. This feels more intuitive. See screenshot below:
screen shot 2019-01-30 at 3 06 11 pm

alyssadelaine:
My changes to the Exclusion List aren't saved -- I add someone to the exclusion list, navigate to the home page where they do not show up, but when I go back to the exclusion list it shows it as being empty.

@kamicut

pages/admin/users-exclusion.js
```diff
@@ -0,0 +1,122 @@
...
+                      Users List
+                    </a>
+                  </Link>
+                </li>
+              </ul>
+            </div>
+            <div className='widget-75'>
+              <div className='row'>
+                <h2 className='header-xlarge'>Add to exclusion list</h2>
+                <p>Exclude mappers from statistics aggregations in the leaderboards.</p>
```
alyssadelaine:
```suggestion
                <p>Exclude mappers from leaderboard statistics and callouts.</p>
```

alyssadelaine:
This works for me now! I just have one request for rewording the "Exclusion List" explanation and then it will be good to go 👍

pages/admin/users-exclusion.js
kamicut: I don't agree with "callout", how about charts? @LanesGood what do you think?

PR #289, opened 2019-02-15
[WIP] Feature/print badges

LanesGood:
The "Print Earned Badges Certificate" should be merged when issues with Firefox and other browsers are resolved.

The badges certificate allows users to print out all their earned badges. Below, the mockup of the badges certificate:
certificatedraft_all

Currently, Chrome users can click a new "View certificate" link in their dashboard to view the page /printallbadges. A button opens the print system dialog, as seen here:
screen shot 2019-02-15 at 1 37 07 pm

**Current issues:**
- There are multiple display issues on FireFox - @page {size:landscape} lacks support, the logo issue is low-resolution, and the page URL headers and footers remain visible. Screenshot here:
[FirefoxScoreboardCertificate_2-15.pdf](https://github.com/developmentseed/scoreboard/files/2870411/FirefoxScoreboardCertificate_2-15.pdf)

- if a user has 10+ badges, the certificate should display as one box on screen, but on print each 10 badges should be a new page with the green border rules all around
- Flex box and page-break after don’t play well together, causing strange padding and layout shifts when I attempt to add additional badges as a test. I need to populate my user data with 10+ badges or test this on a user with 10+ badges 
- @page rules should be set for just for the PrintBadges component/page, rather than all pages. Currently, these print styles apply to all pages. If a user wants to print a page from elsewhere in the Scoreboard platform, these print styles distort the normal view


LanesGood:
Closing this, replaced by WIP PR #290

Issue #255, opened 2019-02-01
Users Can Print a Badge Certificate 


LanesGood:
As proposed by client: when a user achieves a badge on Scoreboard they should be able to print a certificate showing their achievement. Certificate may also be printed by Admin, and can show all badges a user has earned.

Proposed implementation: a button below the "Earned Badges" section that generates the HTML of one of certificate options below, with a print media CSS stylesheet.

![certificatedraft1](https://user-images.githubusercontent.com/12634024/52135009-dec17d00-2612-11e9-863f-76fea0933062.png)
![certificatedraft2](https://user-images.githubusercontent.com/12634024/52135088-0b759480-2613-11e9-9a65-9c6c68b77185.png)

alyssadelaine:
@LanesGood 

Can we remove "Signed" from the second option, and make the background white? We may also want to remove the line under <Name> so we don't need to worry about it looking strange for longer names.

On the first option, let's remove the other badges; there may be too many to list

LanesGood:
@alyssadelaine I thought the "all badges" option was discussed as a client request. I suggest we include a version with and one without. The "all badges" could also be it's own separate printout

LanesGood:
@alyssadelaine see the attached for your review/sending for approval

![certificatedraft2](https://user-images.githubusercontent.com/12634024/52236853-ff4d3980-2895-11e9-8bd5-cbc046f72d6d.png)
![certificatedraft1](https://user-images.githubusercontent.com/12634024/52236855-ff4d3980-2895-11e9-8ecb-124fdcabafe3.png)
![certificatedraft3](https://user-images.githubusercontent.com/12634024/52236856-ff4d3980-2895-11e9-8aee-333777d9a2b7.png)

alyssadelaine:
Oh right @LanesGood I was thinking all badges as separate -- I like the way you have it here

PR #290, opened 2019-02-18
[wip] Printing via pdfkit

kamicut:
Printing using PDFKit working towards #255


kamicut:
@LanesGood back to you, I fixed the page generation

kamicut:
@alyssadelaine we definitely tested by making a loop over 30 badges, but then removing that portion of the code before finishing up the PR.

alyssadelaine:
> @alyssadelaine we definitely tested by making a loop over 30 badges, but then removing that portion of the code before finishing up the PR.

Okay great, thanks! Just wanted to check.

Can you also update the changelog?

LanesGood:
@alyssadelaine Updated with your suggestions - 100% agreed and quite helpful.

Screen Shot 2019-03-26 at 12 37 01 PM

alyssadelaine:
Great, thanks @LanesGood ! I think once the Changelog is updated this will be good to go
cc @kamicut

Issue #286, opened 2019-02-13
Language around recent edits is confusing


alyssadelaine:
We need to standardize and simplify how we're reporting last edit times.

There's a difference of about a day between how the AllUsersTable reports last edit times and how the user's individual page reports the last edit time, and it looks like it's because the former compares with UTC while the latter uses the current time zone. 

We should make all last edit calculations with the current time zone, and in cases where the last edit was "today", it should read "Last edited today" rather than "Last edited <number of hours since midnight> ago."

PR #291, opened 2019-02-21
Fix/report edit times

alyssadelaine:
Fixes https://github.com/developmentseed/scoreboard/issues/286


alyssadelaine:
Now the two edit times are "today" and "2 days ago" so I need to do a bit more work to make the reporting sensical

PR #292, opened 2019-02-21
Remove MVT artifacts

kamicut:
OSMesa produces a __sequences__ layer for housekeeping.
react-leaflet-vectorgrid seems to display these even if there is no
style func associated with that layer. We set the style func for the
sequences layer to 0 opacity


alyssadelaine:
@kamicut Can you update the Changelog?

PR #297, opened 2019-02-26
Feature/add status

### LINKED ISSUE #322: Display "Last refreshed" status

alyssadelaine:
Closes #322 
## Adding OSMESA status to Stats Pages

- [x] All users page
- [x] Individual user page 
    **design question**: current styling results in weird placement
- [x] All campaigns page
- [x] Individual campaign page 
    **design question**: upper right hand corner is currently used for "favorite" and "contribute"
- [x] All countries page
- [x] Individual country page

## Current implementation

Individual User page header:
Individual User Page

All Users page header:
All Users Page

## Questions
- [ ] Is there a more streamlined way to poll the OSMESA api than in each /routes call the way I'm doing here?
- [x] Should the status portion of each header be imported into each page from a single component as opposed to duplicated? If not the whole component, then should the time description at least be imported?
- [x] Does the mock OSMESA API data need to be pulled from a separate file? Is it cleaner that way?


alyssadelaine:
✅ Checked that this worked on staging

alyssadelaine:
✅ Tested on staging

PR #308, opened 2019-03-11
All campaigns in dashboard

kamicut:
This PR:

- Fixes the seeds so that you don't need to run clocks anymore, `yarn seed` will run both the tm clock and user clock in development and prefill the database with data
- Creates an `allCampaigns` item in the user API response for the dashboard. Campaigns are joined with the osmesa hashtags to produce `allCampaigns` which are campaigns that a user has worked on.
- Displays the allCampaigns in the Campaigns table instead of the concatenation of assignments and favorites.

**Next steps**:
@LanesGood we should design the table for allCampaigns to have a fixed height and be scrollable with more rows. Let me know if you need running this PR

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-node_modules
 .DS_Store
 yarn.lock
 package-lock.json
@@ -12,3 +11,7 @@ users.csv
 .vscode
 docs-build
 .swp
+
+# ignore node_modules symlink
+node_modules
+node_modules.nosync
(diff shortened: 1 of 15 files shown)


kamicut:
All Campaigns Widget:

image

Top campaigns widget:
image

These are now out of sync. Should the Top campaigns widget be renamed to "top hashtags"? The All Campaigns widget shows up on the dashboard but not the user profile. The top hashtags show on the user profile and dashboard, but don't have associated names. 

Two paths:
1. Rename Top Campaigns to Top Hashtags
2. Merge the names for campaigns and hashtags in the "Top Campaigns" widget

PR #332, opened 2019-03-26
Fix/empty campaigns

kamicut:
While I think there should be a better overall interface between external services and scoreboard rendering, I think this is a good medium fix.


alyssadelaine:
@kamicut In cases where we can't retrieve any stats, could we change the blurb to something like `No statistics available for this campaign`? And could we hide the table?

PR #340, opened 2019-04-12
Environment variable for "Not Logged In" message

alyssadelaine:
Parameterize Editor name when displaying the Not Logged In message


kamicut:
@alyssadelaine could you create a ticket for custom deployments that are not in openstreetmap: I think we should have a README.md for custom deployments where we include "PROJECT_NAME" and "EDITOR_NAME"

Issue #335, opened 2019-04-01
"Assigned by" column is confusing when all campaigns are displayed


alyssadelaine:
Showing campaigns I participated in as "Assigned by [me]" is confusing. 

Here is the view of all campaigns contributed in with a set of test data:

![image](https://user-images.githubusercontent.com/6155420/55339706-33497300-5471-11e9-8bb4-07232e8106d4.png)

And here is the list of campaigns which were actually favorite-d:
![image](https://user-images.githubusercontent.com/6155420/55339834-73105a80-5471-11e9-9159-25021d8cecc3.png)

I am open to other suggestions, but one option would be to remove the "Assigned by" column and replace it with "Status", which could read "Participated" or "Assigned by [assigner]"

LanesGood:
@alyssadelaine and I discussed and have expanded this issue. Beyond the confusion of the "Assigned by" with the user's own name, there should be a table tab with the user's own campaign contributions highlighted. We arrived at the following solution:

New User Campaigns table tabs:
Assigned | Favorites | Contributions | All


New "All" table column headers:
Name | Sources | Priority

LanesGood:
@alyssadelaine I've created a draft PR with the initial commit. I did not successfully add a new column in all with "sources" referring to the campaign's parent category. Blockers = javascript 

Additionally I discussed with @kamicut, and his thoughts are:
- Drop "all" as a category
- Make assignments the default tab
- Open separate ticket for a notification center/mechanism to surface the projects that are "new" - assigned but not contributed to yet.

I think all of this is getting into an area that would be best to validate with our partners.

alyssadelaine:
Hey @LanesGood , remember that we talked about dropping all as a category, but decided that it would be best to leave it as the default so that if a user receives a new team assignment or adds a favorite, they will see that change reflected immediately in their dashboard.

I don't think we need the notification center option yet, though you can feel free to open a ticket for it.

This was a request from me because the old set of columns didn't make sense once we expanded the data to be "all campaigns". Could you and @kamicut pair to finish this up tomorrow?

LanesGood:
Implemented in #342, closing

PR #342, opened 2019-04-15
Fix/campaign table labels

LanesGood:
Addresses Issue #335 - still need to make some decisions on logic/data for tables


components/dashboard/DashboardAssignments.js
```diff
@@ -23,27 +23,59 @@ class DashboardAssignments extends Component {
     const assignmentFilters = [
       { name: 'Teams', id: 'teams' },
       { name: 'Favorites', id: 'favorites' },
-      { name: 'All', id: 'all' }
+      { name: 'All', id: 'userContributions' }
```
LanesGood: @alyssadelaine and @kamicut The 'all' tab's id has been updated to 'userContributions.' If the campaigns in the user object "all campaigns" array can be filtered to exclude favorites and team assignments, then we can update this tab to "Contributions" and add "all" as a tab that collates all three categories. However, the issue at the moment is that it seems that favorites and assignments also live in the user object "all campaigns" array coming from OSMESA, and that the 'all' tab overrides the selection of the sub-categories.

components/dashboard/DashboardAssignments.js
```diff
@@ -23,27 +23,59 @@ class DashboardAssignments extends Component {
...
       return {
         priority: task.team_priority ? `team priority: ${task.team_priority}` : task.priority,
         name: task.name,
         assigned_by: task.team_name,
-        campaign_hashtag: task.campaign_hashtag
+        campaign_hashtag: task.campaign_hashtag,
+        source: 'ASSIGNMENT'
+      }
+    })
+    let userFavorites = favorites.map(favorite => {
```
LanesGood: @alyssadelaine @kamicut more of a question on object mapping; it seems like I shouldn't have to list out every key value pair for each object just to add a new pair, 'source.' I'm doing this both here in 'userFavorites' and in 'userContributions.' Is there a simpler or better way to do this?

components/dashboard/DashboardAssignments.js
```diff
@@ -23,27 +23,59 @@ class DashboardAssignments extends Component {
...
     })
     const allCampaigns = {
-      favorites: sortBy(prop('priority'), favorites),
+      favorites: sortBy(prop('priority'), userFavorites),
       teams: teamAssignments,
-      all
+      userContributions
     }
+    console.log(userContributions)
     const assignmentsTable = allCampaigns[this.state.assignmentsFilter].map((assignment) => {
```
LanesGood: @alyssadelaine @kamicut Something strange is happening with the tabs and the mapping function here for the `assigned_by` conditional. If you check out your All Campaigns table on your dashboard, as you switch back and forth between the 'Favorites' and 'All' tabs, you'll see that your favorites continue to be mapped/duplicated in both tabs.

components/dashboard/DashboardAssignments.js
alyssadelaine: @LanesGood You're not listing out the key:value pair; you're filtering the original array to create _an entirely new one_.

The map function says, "Take this array and change it in the following way: [...]" so that's what you're doing for each of these arrays before saving the selected items to the new one. 

Your code doesn't know that this is the group of team assignments until you tell it that, in the same way that it doesn't know the source until you define it.

Another way to do this might be to add each of the array items to the corresponding new array within the map function, but @kamicut will have to correct me if that's not very JS.

components/dashboard/DashboardAssignments.js
```diff
@@ -19,36 +19,70 @@ class DashboardAssignments extends Component {
...
+        id: campaign.id,
+        instructions: campaign.instructions,
+        name: campaign.name,
+        priority: campaign.priority,
+        status: campaign.status,
+        tasker_id: campaign.tasker_id,
+        tm_id: campaign.tm_id,
+        updated_at: campaign.updated_at,
+        validated: campaign.validated,
+        source: 'CONTRIBUTION'
```
alyssadelaine:
```suggestion
        source: 'CONTRIBUTIONS'
```

Issue #345, opened 2019-04-16
Some statistics are left out of "summableStats"


alyssadelaine:
These are the [summable statistics](https://github.com/developmentseed/scoreboard/blob/fa63ff24337709b3780174c3872c3437e4060fee/lib/utils/sum_edits.js) used for the "Features Mapped" part of the Campaign header and to calculate the number of user's edits from the data received from OSMESA:

```
const summable = [
  'buildings_add',
  'buildings_mod',
  'poi_add',
  'roads_add',
  'roads_mod',
  'coastlines_add',
  'coastlines_mod',
  'waterways_add'
]
```

Why are `waterways_mod` and `poi_mod` excluded?

PR #346, opened 2019-04-16
Consistent verbiage of edits vs. changesets

alyssadelaine:
New header stats on the user page:

image

And on the dashboard:

image

I also found that the [summableStatistics function](https://github.com/developmentseed/scoreboard/blob/fa63ff24337709b3780174c3872c3437e4060fee/lib/utils/sum_edits.js) is used to get the number of user edits from the OSMESA data, and to report the "Features Mapped" on Campaigns pages. It leaves out `poi_mod` and `waterways_mod`, so [we may need to update that](https://github.com/developmentseed/scoreboard/issues/345).


alyssadelaine:
Note that sum_edits.js is currently in `lib/utils` as well as `api/src/utils/` so ideally that duplication would be avoided by only using it on the backend

api/src/utils/sum_edits.js
```diff
@@ -24,6 +24,12 @@ const summable = [
...
  */
 module.exports = records => {
-  const summableProperties = map(x => Number(x))(reject(isNil, (props(summable, records))))
+  const summableProperties = map(x => {
+    let num = Number(x)
+    if (Number.isNaN(num)) { // control for NaNs in response
+      num = 0
+    }
+    return num
+  })(props(summable, records))
```
kamicut: This function can be shortened, effectively it's `x => ~~x`

alyssadelaine: Thanks @kamicut ! Implemented

PR #352, opened 2019-05-01
Feature/teams update

kamicut:
- [x] Add in database & models for access_tokens
- [x] Add simple oauth service
- [x] Add routes to interact with oauth service
- [x] Add buttons to trigger integration flow
- [x] Gray out UI when oauth service is not available / user does not have access token yet
- [x] Test full flow 
- [x] More tests


sethvincent:
I created a new team, then clicked that team in the list, and got this error:

Screen Shot 2019-05-28 at 5 25 26 PM

Took a look through the teams pages and I'm not sure what's causing this. I bet it's because I didn't add someone to the team yet?

.env.sample
```diff
@@ -8,4 +8,7 @@ OSM_CONSUMER_SECRET=your-oauth-secre
 OSM_DOMAIN=https://www.openstreetmap.org
 SESSION_SECRET=super-secret-sessions
 DATABASE_URL=postgres://scoreboard:test@localhost:5433/scoreboard
-OSM_TEAMS_SERVICE=http://localhost:3000
\ No newline at end of file
+OSM_TEAMS_SERVICE=http://localhost:8989
+OSM_TEAMS_SERVICE_TOKEN_URL=http://localhost:4445
```
sethvincent:
```suggestion
OSM_TEAMS_SERVICE_TOKEN_URL=http://localhost:4444
```

It looks like the port should be `4444` by default. At least that's what got things working for me

.env.sample
```diff
@@ -8,4 +8,7 @@ OSM_CONSUMER_SECRET=your-oauth-secre
...
+OSM_TEAMS_CONSUMER_KEY=teams-oauth-key
```
sethvincent: It would be convenient for `OSM_TEAMS_CONSUMER_KEY` instead be named `OSM_TEAMS_CLIENT_ID`.
Same for `OSM_TEAMS_SECRET`, it would instead be `OSM_TEAMS_CLIENT_SECRET`.

Keeping the same naming scheme as the issuing app would help decrease second-guessing about whether the implementer is setting the right variables.

.env.sample
kamicut: Yes you're correct

kamicut:
Interesting @sethvincent. I got a different error, but pointing to the same page: `this.props.campaigns` is undefined.

kamicut:
@sethvincent I think I've fixed it, something related to a reducer i changed in another branch.

sethvincent:
New error now when clicking a team in the teams list:

Screen Shot 2019-05-29 at 4 42 29 PM

sethvincent:
Not sure if passing that `changePage` prop in `AdminUsersSearch` is strictly required. Maybe we should check if it exists before calling it?

kamicut:
Yeah, this is related to the new reducer introduced by merging #363

sethvincent:
Can now successfully manage teams!

PR #355, opened 2019-05-09
Fix/235 campaign navigation

### LINKED ISSUE #235: Navigating back from individual campaign page to Campaigns always takes you to the first page
If I'm on the fifth page of Campaigns and I click an individual campaign, I want my browser's "back" button to take me back to the fifth page, not the first

kamicut:
This fixes #235 

While we have a redux store (using unistore), we never take advantage of the client side routing, so the state gets reset when we navigate from campaigns to campaign because we call an explicit SSR every time through the express server.

- [x] Fix client side routing for campaigns
- [x] Fix filter components to read from state
- [x] Test pagination

### DIFF
--- a/components/Link.js
+++ b/components/Link.js
@@ -4,8 +4,12 @@ import { APP_URL_PREFIX } from '../api/src/config'
 
 export default class NewLink extends Link {
   render () {
-    const { href } = this.props
+    const { href, as } = this.props
     const newHref = join(APP_URL_PREFIX, href)
-    return <Link {...this.props} href={newHref} />
+    let newAs
+    if (as) {
+      newAs = join(APP_URL_PREFIX, as)
+    }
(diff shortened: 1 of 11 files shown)


alyssadelaine:
Also the pagination should be reset when any of the filters change

kamicut:
@alyssadelaine yes, I think that would be useful. @LanesGood how do you imagine a "Clears Filters" button to look like?

alyssadelaine:
@kamicut Can you go ahead and implement "Clear Filters" with, for example, a link like the one to the Scoreboard Documentation in the footer and then we can circle back on the Design later?

LanesGood:
@kamicut and @alyssadelaine I'm a fan of using a simple `link--normal` style here - it's unobtrusive, and doesn't take attention away from the filters themselves.

PR #366, opened 2019-05-22
Close #320

### LINKED ISSUE #320: Duplicate campaigns
Some campaigns are appearing more than once in the Campaigns panel

kamicut:
- I removed the duplicates manually from staging
- I added a constraint to the database


api/src/services/tm_types/tm2.js
```diff
@@ -115,7 +115,7 @@ class TM2API {
         // not found
         return db('campaigns').insert(obj)
       } else {
-        return db('campaigns').where('tm_id', obj.tm_id).update(obj)
+        return db('campaigns').where('id', rows[0].id).update(obj)
```
alyssadelaine: @kamicut Why is this necessary if the column is constrained to be unique?

kamicut: Because we might have multiple tasking managers, so we select the project depending on tasker_id _and_ tm_id

alyssadelaine: @kamicut My comment here was in reference to `rows[0]`, since there should only be one item returned if the column has a uniqueness constraint, right?

kamicut: ah, because knex always returns an array even for a single row, so it's sure to be the first item in that array.

PR #370, opened 2019-05-28
Fix/deployment docs

kamicut:
Not sure if we should include a note on how to deploy osm-teams service in these deployment docs. Should we have a reference to dependent services like OSMesa, Teams and maybe tasking manager?


kamicut:
@sethvincent will add in new PR.

sethvincent:
:+1:

PR #372, opened 2019-05-29
Don't count excluded users or their edits in top stats on home page

### LINKED ISSUE #369: Investigate difference in edits on home page vs. Users page
My suspicion is that the statistics feeding these pages deal with excluded users differently

alyssadelaine:
Fixes #369 
✅ Confirmed this was the reason edit totals were different on Home page vs. Users page


kamicut:
Thanks!

PR #376, opened 2019-05-30
Hotfix ramda deprecated method

kamicut:
mergeRight is deprecated in new versions of Ramda


alyssadelaine:
The problem this was aimed at was solved by doing a fresh `npm install` on the server

PR #384, opened 2019-06-12
Feature/combined stats

### LINKED ISSUE #383: Add modified stats to campaign page
Add `modified` statistics on the table in the campaign page, so that a more complete picture of contributions to a campaign can be painted.

- There may need to be a change to the blurb at the top of the table to ensure that all stats are represented
- Will need to propagate to the .csv export as well
- Any stats that are currently created+modified should be split out

LanesGood:
Combines the `add` and `mod` stats for user edits on the Campaign page blurb, table and .csv export, the Dashboard blurb and .csv export, and the User profile blurb and .csv export. Closes #383 

It appears that the stats rendered by the Campaign page blurb are static and randomly generated, while the Campaign page blurb on staging/production is an accurate and dynamic sum of the campaign participants. For now, this means that the campaign page blurb is not a sum of the participants stats, so it is not possible to confirm the accuracy of the calculation. @kamicut can you confirm that this is the case?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Updated osm-teams to have access tokens. Admins now need to sign in to osm-teams
 - UI additions to gray out teams if the user does not have an access token
 
+### Changed
+- Stats in the tables, exports and blurbs on Campaign, Dashboard, and User profile pages now combine the `added` and `modified` edits
+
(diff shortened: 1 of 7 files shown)


kamicut:
Thanks for working on this. I'll check to see the accuracy of the development stats, we should lean towards it being predictable so I'll add to your PR if necessary.

kamicut:
I am also moving the changelog entry to Unreleased

LanesGood:
@kamicut agreed, will open a separate issue for that.

PR #399, opened 2019-07-16
Campaign statuses

### LINKED ISSUE #392: Make sure draft campaigns are not displayed
In some instances draft campaigns can appear in the scoreboard campaigns page. We should ensure that draft campaigns are not ingested using the tasking manager clock.

kamicut:
Closes #392 

This makes a few fixes to campaigns so that draft campaigns are never displayed in the campaign search:
1. Make status and priority a text column in the database
2. Convert integer priority and status in the tasking manager clock process for tm2 so that they are strings and match tm3
3. Filter on the string 'DRAFT' on status in the API call for campaigns

I've also added a test for the logic, by randomizing the status in the test seeds and making sure the drafts are filtered out.


package.json
```diff
@@ -84,6 +84,7 @@
     "axios": "^0.19.0",
     "babel-preset-jest": "^23.2.0",
     "body-parser": "^1.18.3",
+    "chance": "^1.0.18",
```
sethvincent: this should probably be a dev dep

PR #409, opened 2019-08-02
Add archived campaign toggle for campaigns page

### LINKED ISSUE #408: Add UI Filter for "Archived" campaigns
For archived projects, it'd be good if there was a toggle to include archived projects similar to how this is done in the tasking manager. As a user, this will allow for streamlined searching and hiding some additional noise in the project lists.

kamicut:
Closes #408

### DIFF
--- a/api/src/db/seeds/development/campaigns.js
+++ b/api/src/db/seeds/development/campaigns.js
@@ -1,5 +1,6 @@
 const yakbak = require('yakbak')
 const path = require('path')
+const chance = require('chance').Chance()
 const http = require('http')
 const tmWorker = require('../../../tm_clock')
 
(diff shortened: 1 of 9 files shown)


LanesGood:
@kamicut Looks good, but there are a few small issues we found:
- the "X out of X campaigns" string doesn't seem to update correctly
- the archived campaigns toggle doesn't filter correctly when combined with the TM filters
- Toggling archived campaigns doesn't trigger the "Reset Filters" link

I can share a screen recording of point 2, but you'll see what I mean if you start playing with the TM filters, the archived campaigns toggle, and the search bar all together.

kamicut:
@LanesGood I think issue 1 and 2 are related. I think I fixed all three issues if you could check!

LanesGood:
@kamicut The archived campaigns toggle looks good. I'll add in some checkbox styling and then pass back to you. 

I still see some issues with the validation + completion sliders, when combined with the `Search` input and the TM filters... this is likely a different issue and should be opened separately.

PR #412, opened 2019-08-15
Add railways to Campaign and User tables/blurbs/export

LanesGood:
This PR adds railways to the Campaign, User and Dashboard blurbs, tables and CSV download in anticipation of the railway stats being added to the OSMesa server.

@kamicut to confirm: should the railway stats be added to the API calls as I've done here?


kamicut:
@LanesGood could you add something to the changelog?

PR #416, opened 2019-08-29
osmesa schema updates

sethvincent:
This pr makes the necessary changes to adopt the new schema from the stats server.

There are a few remaining issues:

I think we need to construct the array of users for the campaign table based on the ids in a campaign's `user_edits` object.

The `extent_uri` needs to be constructed based on the campaign/user id, it's no longer a property on data returned by the stats server.

It seems that sum_edits.js is no longer needed as there is now an `edit_count` property returned by the stats server.

I'm unfamiliar with the `user_country_edits` table in the db and related code. I made a few fixes to get the tests to pass. We should double check all is correct with that.


kamicut:
We are closing this for now. Thanks @sethvincent

Issue #418, opened 2019-09-25
Confirm `last edit` calculation in Scoreboard


LanesGood:
Question from a user asking if the last edit is based on a time zone or if today means within the last 24 hours. 

Code in: https://github.com/developmentseed/scoreboard/pull/291

LanesGood:
This was confirmed, closing.

PR #419, opened 2019-09-25
fix/418 Updates the last edit to reflect current math

necoline:
#418 Currently the time util is calculating 'today' any time within 24hrs of the last edit. This appears to the user as bug when the user is viewing the page on the day following the last edit, but 24hrs has not actually passed since the edit time.

My short solution was to change the text from 'today' to 'In the last 24 hours' 

An alternative approach would be to calculate the time from the user edit time and their time zone to figure out when 'today' ends for them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
+- Edit times that occured within the last 24 hours show 'Last Edit: In the last 24 Hours' rather than 'today'
+
 ## [v1.4.0] - 2019-08-19
 ### Added
 - Campaigns can be filtered to include or exclude archived campaigns
--- a/lib/utils/format.js
+++ b/lib/utils/format.js
@@ -39,9 +39,8 @@ export function formatEditTimeDescription (timeVar) {
   const now = new Date()
   // divide difference in timestamps by millisecond conversion
   const daysDifference = Math.floor((now - timeVar) / (1000 * 60 * 60 * 24))
-  const hoursDifference = Math.floor((now - timeVar) / (1000 * 60 * 60))
   if (daysDifference < 1) {
-    return `${hoursDifference} hours ago`
+    return 'In the last 24 hours'
   } else if (daysDifference === 1) {
     return `1 day ago`
   } else return `${daysDifference} days ago`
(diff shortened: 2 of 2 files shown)


LanesGood:
@necoline thanks for the quick fix! Let's hold on implementation.

One other alternate approach, and probably much easier than figuring out a user's time zone, would be to return the hours since the last edit if in the last day.

P.S. I believe this should be fix/418

necoline:
I like your suggestions much better than the different definition of today

necoline:
To test the hour count, I subbed the arg `timeVar` with `new Date('Tue Sep 25 2019 00:00:00 GMT-1400 (Eastern Daylight Time)')`

LanesGood:
@necoline looks great, and thanks for doing the initial test. If we decide to merge we should also build this into the API tests. If you'd like you can add that into this PR along with a changelog update.

necoline:
Oh yes, I forgot to look for associated tests. Thanks for the nudge.

kamicut:
@necoline, are we effectively reverting https://github.com/developmentseed/scoreboard/commit/b493cb69bc2bce25c2d39b2d42dfeb4d421abd97 ? If so we can probably re-use `distanceInWords` since we have it on the page. I'd ask @alyssadelaine why we did that in the first place.

necoline:
@kamicut I didn't take a look at `distanceInWords` in my first time through since it wasn't actively used for this purpose. It looks like it's an ideal solution. Is there a reason it wasn't being used on this method in the first place?

alyssadelaine:
I don't think it's reverting that commit since this only applies to edits within the last day. 

@kamicut @necoline 

We [originally used](https://github.com/developmentseed/scoreboard/pull/291/files) `distanceInWordsToNow`, which was [deprecated](https://github.com/date-fns/date-fns/issues/377) in favor of `distanceInWords`. 

If you look at the full PR, I [implemented and then took out `distanceInWords`](https://github.com/developmentseed/scoreboard/pull/291/commits/b493cb69bc2bce25c2d39b2d42dfeb4d421abd97) in favor of calculating my own difference in days. I can't remember my reasoning behind that, but if I had to guess, it may be because I thought we would want the precise number of days (with consistent formatting) over the [vaguer](https://date-fns.org/v1.30.1/docs/distanceInWords) descriptions in `distanceInWords`.

In any case, I'm not attached to the solution I chose as long as it doesn't reintroduce the [error we were having in consistency of edit descriptions](https://github.com/developmentseed/scoreboard/issues/286).

necoline:
@kamicut pointed out that the issue with `distanceInWords` is the fact the we don't get full time stamps from OSMesa. The time stamp that is available in `formatEditTimeDescription` is assigned somewhere in one of the header components, meaning that is not accurate. For that reason, I'm going to revert this to my second commit where we just change the text showing for anything less than 1 day.

necoline:
@LanesGood I won't be adding tests since there aren't any tests for the frontend.

PR #424, opened 2019-09-27
Fixes map controls on campaign pages

LanesGood:
Quick fix to enable map controls on campaign pages, which were previously buried in the z-index stack.


kamicut:
@LanesGood thanks! Could you add to the changelog?

PR #425, opened 2019-09-30
Feature/loading state

necoline:
**Description:**
Adds a component loading state that fills the width of its parent component.

**Rationale:**
This is useful for conveying to the user that a feature exists and is forthcoming. 
Currently there were a few empty divs that were unable to communicate this, there were some spans that showed `loading` in which case, this is more beautification than anything.

**Future Work:**
- There are a number of components that could use this that are not included in this pr. I have only changed out loading states where existing logic exists with the exception of the admin components like [this](https://github.com/developmentseed/scoreboard/blob/b1a628c93f1cf4452bcfc1b58f4f635170de17d0/pages/admin/teams-edit.js#L149) one. It would be quite easy to add in a loading state beneath this component but I need advice on how to test that it is working as expected. I don't think my user permissions allow access to those pages. I may be wrong.

- The site could do with a global loading state. This PR does not offer this. I spent some time looking through the codebase to find a place to do implement it. @LanesGood @kamicut mentioned the complexities of addressing this before. Now that I am more familiar with the codebase I would be up for another discussion for how this might be possible. Implementation of the loader would be a simple as adding a global wrapper around the component, the location of this component is tbd.

users

### DIFF
--- a/components/campaigns/CampaignsListing.js
+++ b/components/campaigns/CampaignsListing.js
@@ -1,5 +1,6 @@
 import React from 'react'
 import CampaignCard from './CampaignCard'
+import { LoadingState } from '../common/LoadingState'
 import { range } from 'ramda'
 
 export default ({ records, apiStatus, total, allCount }) => {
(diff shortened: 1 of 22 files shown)


kamicut:
@necoline to test out admin pages, follow the instructions in the README do give yourself "administrator permissions"

necoline:
I added loaders to the admin pages
image

Issue #428, opened 2019-10-09
Common location for labels and definitions across the site


necoline:
If we create external files with header labels and definitions, we can easily apply this information broadly across the site and create the structure to make internationalization of the site far easier in the future.

kamicut:
closed in #429

PR #427, opened 2019-10-08
Feature/380 tooltip

necoline:
#380 
Here is the first iteration of the the tooltip.
Is this the type of tooltip descriptors that you are looking for?
Do we want them for all row headers in the forms?

image

PR #429, opened 2019-10-09
Feature/428 glossary

necoline:
## Description
**Adds a common file containing labels and descriptions for table headers**
Rather than hard coding values into table headers, we can can pull language from a json file for the titles and descriptions.

## Related Issue
#428 

## Motivation and Context
This will help 
- reduce the use repetitive, hard coded values
- aid the process of internationalization of the site
- provide a scaffold for how to organize language for the site
- simplify the process of editing descriptions for the tooltip 

note: this access the tooltip PR ( #427 )

## How Has This Been Tested?
I have tested each of the changes by viewing the changes in each component effected

- the json offers implicit ordering of labels. 
On some level, this consistency is helpful, but there are instances where there is a need to have tables headers ordered in different ways. For example, most of the tables show the name before the user id, but on the admin user list, the table prioritizes the user id ahead of the name. I don't know if there is flexibility around this order, but I assumed that this wasn't the case and I added a sort function after the filter like so:
https://github.com/developmentseed/scoreboard/pull/429/files#diff-866a7a0aeaad83f0cde7b02d165f5b7dR49

image

## Screenshots (if appropriate):
image

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code is readable and well commented.
- [x] My changes are reflected in the changelog.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.


LanesGood:
@necoline One other question: would this glossary also serve to provide tooltips for non-table-header elements? For example, could the glossary provide descriptions for page headings, or inline text? Taking it another step further, could we use the glossary to generate a static page with all definitions?

necoline:
When I started this, I envisioned multiple json files with different categories of text. For example, 
- table headers
- page headers
- key phrases

It might make more sense to organize language by category and pull in those files as they apply like
- users
- campaigns
- countries

It's open for discussion I think there are pros and cons to both. The pro that comes with this setup (as opposed to the previous version) is that all of the language is in one location (lib file) where it can be easily updated, translated, reorganized and repurposed. In the old way, everything was hard coded which is as brittle as it gets. 

To answer your last question, though the current single file with table headers likely needs some finessing and we will need to add more files to our lib, it does set us up well for populating content for a static page.

LanesGood:
@necoline this all sounds good. How do we determine whether the single file is the best way forward?

/cc @kamicut

necoline:
I have an alternative data structure that I think might improve the universal applicability of the glossary. 

It still needs to be tested but it would look something like this #430

necoline:
Sorry I didn't get a chance to test it thoroughly before hopping on that mtg. Your fixes look good 👍

PR #430, opened 2019-10-15
Feature/428 glossary hash

necoline:
- Moves definitions into a common glossary file
- Creates table json file with an id and list of headers (in the order that they should be listed)
- Creates a common util that returns the definitions specified in the selected header and sorts the definitions by the order stated in the table file.

- Applies the selector to all the tables


necoline:
- Adds optional tooltip 
The headers with tooltips are defined in a separate array in `table-header.json`, the selector then adds a tooltip boolean property on the glossary which is used to conditionally display table headers with or without tooltips.

- Makes table headers enums rather than using strings as a selector argument

- Applies dynamically added headers with optional tooltip
To test, I removed tooltips on some of the tables. For example, I omitted a tool tip for all the headers of `name` since I imagine that would not require explanation for the most part. 

Here are some screenshots:

Screen Shot 2019-10-16 at 2 45 09 PM

Screen Shot 2019-10-16 at 2 45 18 PM

Screen Shot 2019-10-16 at 2 52 28 PM

Screen Shot 2019-10-16 at 2 41 43 PM

PR #432, opened 2019-10-21
Feature/osmesadb

kamicut:
Moves statistics to be pulled directly from OSMESA database


kamicut:
@sethvincent I've added `OSMESA_DB` to README and deployment docs

PR #434, opened 2019-10-28
v1.6

kamicut:


LanesGood:
@kamicut does this enable the glossary definitions for all terms? I thought we might wait on these until text is approved.

kamicut:
@LanesGood I fixed some of the text, you can check staging?

LanesGood:
@kamicut let's remove the definitions from:
- Rank
- Name
- POI

I'm not yet familiar with the quick way to turn these off and on. Right now I think it's in "Verbose mode" which is good for staging, but a bit over the top on production IMO

PR #436, opened 2019-10-31
Replace `edit_sum` with new `edit_count` stat for users and dashboard

LanesGood:
Fixes discrepancy between user total edits displayed on /users and the individual user/dashboard page. The later was still relying on `edit_sum` calculated internally before the schema change. The new schema includes an `edit_count` with all user edits served directly from OSMesa. 

The `sum_edits.js` util is still needed by campaigns to collate and sum all user edits on an individual campaign. However, we may want to explore removing this if possible as well.


pages/dashboard.js
```diff
@@ -66,6 +66,7 @@ class Dashboard extends Component {
 
     const { badges, teams, refreshDate } = account
     const osmesaData = account.records
+    console.log(osmesaData)
```
kamicut: Please remove this console.log
```suggestion
```

pages/user.js
```diff
@@ -41,6 +41,7 @@ export class User extends Component {
       teams,
       refreshDate
     } = this.props.user
+    console.log(records)
```
kamicut: Please remove this console.log
```suggestion
```

LanesGood:
@kamicut yes, the `edit_sum` is still being used in the campaign page to sum up all contributors' edits. (see original comment)

kamicut:
Ah, ok! I think there is an edit_count for campaign pages as well. I can take a look at this.

LanesGood:
@kamicut looking at the campaign props, I see `stats` with all of that campaign's top-level feature stats included. The campaign `stats` also include an array of `users` with an `edit_count` that looks specific to this campaign. It will be an easy fix to replace the `editSum` currently filling the "Total Features Mapped" cell in the Campaign UI with either A) the total edits recalculated from the campaign's aggregate stats, or B) the summed `edit_count` of the users captured by the campaign's stats.

What's interesting for both edit_sum and edit_count is that they are capturing all features added, modified, and deleted. So, the "Total features mapped" language is a bit misleading; it's really "Total Edits." Should we make this change?

LanesGood:
@kamicut on further investigation, I found a discrepancy between the sum of all a campaign's users' `edit_count` properties, and the summed features (including _add, _mod, _del) for a campaign's top level stats. The summed `edit_count` is consistently greater than the summed array of features.

I also see that campaign stats do _not_ include railways as a feature. I'll open a separate issue for this. I'm not sure if this is on our side or coming from OSMesa.

With some number crunching I now see a bigger issue with the original discrepancy raised. The new `edit_count` property for users is always greater than `edit_sum`, and includes things we hadn't originally included in the "summable properties" used for `edit_sum.` If we want to continue to make the distinction between "Total Features Mapped" and "Total Edits" we should clarify what these stats mean. I'm assuming that a "Total Edits" includes the OSM definition of an edit: any node added/modified/deleted on OSM; whether a node on a way, or a field for a POI, etc. I think the "Total Features Mapped" stat is a bit more interesting for a rolled-up view, for example a country or campaign. However, perhaps we stick with the new `edit_count` for campaigns/countries, and adding a "Totals" line for campaigns/countries can provide the necessary detail on total features mapped...

With the new `edit_count` property, a user's total edits would reflect all nodes they've touched. However, the "Type of Edits" piechart on the dashboard/profile page is only displaying the edits captured by the "summable edits" property. The sum of the types totals in this piechart is currently equal to the `edit_sum` value rendered on the user's page.

With all of this in mind: what do you think is the most useful stat to surface on the user's page, and throughout the app? My opinion: let's switch everything to the totals coming from OSMesa for simplicity's sake, but note the definition of "Edits." Then, on the CSV download, include the missing details: _add _mod and _del for all features, as well as all distances.

kamicut:
I agree with your assessment. I think we should have a single source of truth, and `edit_sum` existed before the schema changes in OSMesa. We should remove any calculation we are making on the frontend and try as much as possible to fall on properties given to us by the database. With that in mind, what do we need to do for this PR?

LanesGood:
@kamicut we just have to remove `editSum` from the campaign page, and replace this with the summed `edit_count` from all users in the campaign. I see that we're doing the editSum on the API call - how would we do this here?

LanesGood:
@kamicut ready for final review. I believe the reduce that I've used on the api util `api/src/utils/sum_editCounts.js` could actually be inlined into the route, just like in `api/src/routes/countries.js` - however, my attempts at this weren't working. If you'd like to correct this please feel free; if it's fine as is, we can merge.

PR #441, opened 2019-11-06
Feature/438 sortable table

kamicut:
This allows the tables in the campaign page to be sorted by clicking on the header. Please try it out! You'll need to `npm install` to get `react-table`

@necoline what do you think of this implementation for the table headers?
@LanesGood could you check the design?


components/campaign/CampaignTable.js
```diff
@@ -20,37 +24,103 @@ export default function CampaignTable (props) {
       km_coastlines_add_mod: user.km_coastlines_add + user.km_coastlines_mod,
       km_waterways_add_mod: user.km_waterways_add + user.km_waterways_mod
     }))
+
+  const headerDivs = prepareAllHeaders(tableHeaderNames.CAMPAIGN)
+  const formattedNum = ({ cell: { value } }) => formatDecimal(value)
+  const { headers, rows, prepareRow } = useTable({
+    columns: [
```
necoline: This seems to subvert our initial goal of avoiding hardcoded values in the tables. I would try to consolidate this as much as possible so we can avoid manually creating each table

components/campaign/CampaignTable.js
```diff
@@ -20,37 +24,103 @@ export default function CampaignTable (props) {
...
   return (
     <div className='widget clearfix table-wrapper'>
       <table>
         <thead>
           <tr>
-            <TableHeaders tableName={tableHeaderNames.CAMPAIGN} />
+            {
+              headers.map(column =>
+                <th {...column.getHeaderProps(column.getSortByToggleProps())}>
+                  {column.Header}
```
necoline: Why not use the react-table sortMethod in the column array above?

LanesGood:
I also see that by mapping `prepareAllHeaders` into new `<th>` tags, we're creating a redundant nested `<th>`
image

So it looks like we can't use the two components independently, but need to somehow combine them.

What do you think of the table set up in the [Af-Elections project](https://github.com/nditech/af-elections/blob/develop/app/assets/scripts/components/elections/results/results-table.js)? There are a lot of components involved, but it looks like a good way to have the ability to configure sorting, tooltips, and even table cell alignment for each instance of a table.

components/campaign/CampaignTable.js
kamicut: I'm not sure what you're referring to?

components/campaign/CampaignTable.js
kamicut: One idea could be to turn the TableHeaders tooltip component into a react-table custom hook, and so we could have the logic for the tooltip happen within the react-table rendering. We could then have a custom Table component that reads the glossary and has certain react-table hooks to generate the columns. 

The problem I had with just generating the column names is that the accessor methods to the data and the glossary ids are different. We can standardize them somewhere but I'm wondering how to do it in a way that's not brittle.

kamicut:
@necoline I agree with your comment about manual entry of columns. However the previous code had us explicitly state accessor methods in the row rendering, which amounts to the same thing. We could of course abstract everything into a react-table component and have the accessor logic ripple from the table headers json. Does that mean that we make this PR a full refactor of the table components across the application? Happy to move forward with that if @LanesGood agrees. 

@LanesGood I removed the rank because I didn't see it make sense when the tables are sorted by different columns. What do you think? I'll also try to make the chevron and tooltip render within the same element.

LanesGood:
@kamicut I think a refactor of the tables is a good idea. Sortable columns will be a useful enhancement anywhere, as will removing manual column entry. Hopefully this will play well with the CSV export library.

I agree that removing rank in this table makes sense. We should have the table sorted by most edits by default, which gives the same info. We may want to look into removing rank from everywhere except the users and teams tables.

kamicut:
@necoline I've created an abstract component based on what we've discussed on Thursday. It contains a schema for each table that you pass through the component and headers will be created using react-table. 

If you think this is a good approach I could do the same thing for all tables.

@LanesGood I fixed the duplicate `<th>` which fixes the span issue. Please check it out.

components/campaign/CampaignTable.js
```diff
@@ -1,15 +1,51 @@
 import React from 'react'
-import Link from '../Link'
 import { sortBy, prop } from 'ramda'
-import { formatDecimal } from '../../lib/utils/format'
 import CSVExport from '../../components/CSVExport'
-import TableHeaders from '../common/TableHeaders'
-import { tableHeaderNames } from '../../lib/enums'
+import Table from '../common/Table'
+
+const tableSchema = {
```
necoline: Moving forward, do you think we should keep the tableSchema on the component level? If we think that is the way to go ( something I'm leaning towards), would it make sense to include `columnOrder` and `displayTooltip` attributes on the headers object?

components/common/Table.js
```diff
@@ -0,0 +1,114 @@
+import React, { useEffect } from 'react'
+import Tooltip from './Tooltip'
+import Link from '../Link'
+import { useTable, useSortBy } from 'react-table'
+import { toPairs } from 'ramda'
+import { formatDecimal } from '../../lib/utils/format'
+const glossary = require('../../lib/i18n/glossary_en.json')
+const formattedNum = ({ cell: { value } }) => formatDecimal(value)
+
+function chooseRenderer (datatype, idMap) {
```
necoline: I would probably rename this to something like selectCellFormater or something else a little more specific

components/common/TableHeaders.js
```diff
@@ -1,9 +1,14 @@
 import React from 'react'
 import Tooltip from './Tooltip'
+const { values } = require('ramda')
 const tableHeaders = require('../../lib/i18n/table-headers.json')
 const glossary = require('../../lib/i18n/glossary_en.json')
 
 const TableHeaders = ({ tableName }) => {
+  return values(prepareAllHeaders(tableName))
```
necoline: Shall we revert changes in this file since they aren't being used with your recent commits?

components/campaign/CampaignTable.js
kamicut: I was thinking that displayTooltip could be either:
1. as an array, it could be less repetitive than adding a true false to each header
2. as an object, it makes sense to be a property on each header

I couldn't figure how the column order could be collapsed as a property on the headers? would it be a numeric ranking? I think an array makes sense because it has an order (the index of the element)

components/common/TableHeaders.js
kamicut: I think if we go ahead and do all the tables using the Table component, then we don't need the TableHeaders component or table-headers.json because they would be handled a) by the schema in each component b) by the Table component.

necoline: I'm with you there

components/campaign/CampaignTable.js
necoline: If the schema is defined on the component level we could leave the order of the objects as the implicit order of the headers, otherwise I was thinking a numeric value for the order. 

With this new setup, I think it makes sense to add the tooltip as a property on the table header object

LanesGood:
@kamicut looking good!

LanesGood:
@kamicut and @necoline I made multiple style changes here that also required changing the structure a bit. This included setting the entire <th> contents inside an <a> tag to:
1) allow users to sort a column by clicking anywhere on the column header, not just the icon, and 
2) allow the inclusion of two :after elements in one parent

LanesGood:
@kamicut looks great! 
My styling introduced a bug that I haven't figured out - the icons don't always load correctly, and are positioned behind the text like the below. Refreshing always fixes this. Is this a server side rendering issue?
image

One minor bug found is that the "sort by" input in the Countries table no longer works; the same input in the main Users table does still work.

Finally, the assignments table also looks like it's throwing this error for the dashboard:
`|   const campaignMap = Object.assign(...assignments.map(({ name, tm_id, tasker_id }) => ({ [name]: `${tasker_id}-${tm_id}` })))`

components/AssignmentsTable.js
```diff
@@ -1,41 +1,28 @@
 import React from 'react'
-import Link from './Link'
-import TableHeaders from './common/TableHeaders'
+import Table from './common/Table'
+
+const tableSchema = {
+  'headers': {
```
necoline: Since we are declaring the schema on the component, would it make more sense to move the display toolTip in this object as a boolean make the order or properties the implicit column order?

kamicut: We decided to do this in a future refactor

LanesGood:
@kamicut in addition to the campaignMap empty object bug, I'm seeing this in my dashboard assignments table:
image

Issue #443, opened 2019-11-11
Campaigns that only have modified data will fail to display data


kamicut:
We don't guard for the case where there is only modified data in the campaign. Campaign tables try to sum up added + modified statistics, but if there is no "added" property, the sum fails.

LanesGood:
Closed by #444 and #455

PR #444, opened 2019-11-11
Fix defaults for OSMesa stats

kamicut:
This fixes the defaults for stats for #443


kamicut:
I haven't had that problem when switching branches? Let me know if this happens again

PR #447, opened 2019-11-11
Fix edit user link

### LINKED ISSUE #446: Scoreboard Edit Profile Link
When selecting "edit profile" from user menu, an additional {{scoreboard }} is added to the link

LanesGood:
@kamicut I believe the `join(APP_URL_PREFIX)` is not needed here, and causing a bug in contained environments. Is there a way we can test this on a live environment without deploying?

Fixes #446


kamicut:
That's a good question, I think we'll have to test it on staging because setting a URL prefix tester locally is a much bigger lift. Maybe that could be a new issue?

PR #451, opened 2019-11-27
Feature/total-row

### LINKED ISSUE #442: Add Totals line for campaign tables

LanesGood:
Starting PR to close #442 

Feature Requirements:
- [ ] sum up all cells in a column
- [ ] only apply the sum function to a column with `number` datatype
- [ ] Add "Total" as a row label (& remove link to non-working userpage)
- [ ] [Enhancement] Make "Total" row the first row in the table
- [ ] [Enhancement] Style "Total" row differently to highlight totals
- [ ] [Enhancement] Have "Total" row respect table sorting 
- [ ] [Enhancement] right-align all `number` table cells

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 
 ## Unreleased
+### Added
+- Add country of origin column to campaign tables
+
 ### Changed
 - Refactor tables and tableheaders to use a common component configured using a tableSchema
 - Make all tables sortable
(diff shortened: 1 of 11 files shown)


LanesGood:
@kamicut a few questions:
1. Apologies for the crazy rebase - I attempted to use GitKraken for this and obviously failed... and don't want to mess this up further. Strangely, the commit history looks clean on the branch, but not above... any suggestions?
2. I've added the totals as an additional item in the `campaignTopStats` object. However, I think they should instead show up in the table footer, so that they can be styled separately and not affected by the table sort.
3. To do this, I suspect I need to use the react-table `aggregator` hook, but am a bit stumped by this. Working on trying to understand it, or to instead use the current totals as a footer object passed to the general Table.js setup, which may be simpler. What do you think?

LanesGood:
Closing in favor of #459 to avoid the above commit soup

PR #452, opened 2019-11-27
Add country of origin to campaign tables

kamicut:


LanesGood:
@kamicut BTW I'm wondering if we need to avoid printing 'N/A' for certain fields - not just in the country column for the campaign table... but this can be a separate issue, and may not actually be a problem.

PR #453, opened 2019-11-30
Fix/449 empty tables

### LINKED ISSUE #449: Table Pagination lists empty pages
Even with 0 table rows, all paginated table list many more pages than there are actual rows:
image

Users can navigate to these empty pages:
image

The max number of pages seems to be set at the max number of items. Table pagination should reflect sorted and filtered tables, and only display numbers for the pages of actively filled items.

kamicut:
This fixes #449 for the users table. @LanesGood is there another table that has pagination that has this bug? The countries table doesn't have pagination, and the campaigns pagination doesn't have this problem.


LanesGood:
@kamicut Confirming that it doesn't seem that this same issue exists locally for the campaigns page, or in production instances. No, no other paginated table seems to have this issue.

PR #459, opened 2019-12-17
Sum data in table footers

### LINKED ISSUE #442: Add Totals line for campaign tables

LanesGood:
Total row in campaign tables, closes #442. Replaces #451


LanesGood:
@kamicut Circle CI fails on a module and says "The engine "node" is incompatible with this module." I think this is an error. Can you restart the test when reviewing this PR? (I don't have admin permissions to do so)

Issue #344, opened 2019-04-16
Add instructions to README for creating admin in dev environment


alyssadelaine:
We made a few changes which make the [Create an Admin Instructions](https://github.com/developmentseed/scoreboard#create-an-admin-user) less straightforward in development. We should amend the README to explain the new instructions.

LanesGood:
Closing in favor of #316

Issue #316, opened 2019-03-13
Update the README


alyssadelaine:
The README isn't very easy to read, particularly in the way `Development` and `Production` are divided even though most of the steps overlap. Some sections and commands also seem to have fallen out of date, such as adding an admin by username

LanesGood:
From #341 - 
> we should have a README.md for custom deployments where we explain "PROJECT_NAME" and "EDITOR_NAME"

LanesGood:
From #344 -
> We made a few changes which make the Create an Admin Instructions less straightforward in development. We should amend the README to explain the new instructions.

PR #466, opened 2019-12-20
Update README.md

LanesGood:
Include --osm-id and link to http://whosthat.osmz.ru/ in instructions for creating an admin user

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,10 +103,10 @@ To be able to develop admin pages you'll need to have an admin user.
 Make your user an admin by running:
 
 ```
-./api/bin/scoreboard create-admin-user --username {your osm username}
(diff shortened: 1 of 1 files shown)


kamicut:
Not sure why we need this @LanesGood? I can add myself as an admin with a username and it works. Are you facing an issue here?

LanesGood:
@kamicut this fixes #344 and contributes to #316. I haven't been able to add myself as an admin in local development with just username since well before the formerly mentioned ticket's creation; not sure why this would be different for you unless there are legacy config files on your local machine.

kamicut:
This is a deeper issue that I don't think should be a README fix.

PR #467, opened 2019-12-20
Dashboard blank state

LanesGood:
This PR adds a blank state for the user dashboard, and a null message for the dashboard blurb. 

If a user has not made any edits, the blurb directs them to the campaign page.

If a user has not created any edits and they are not an admin, the page renders with a background image:
Screen Shot 2019-12-20 at 7 48 07 PM


LanesGood:
@kamicut this is now fixed for admin views:
image

PR #468, opened 2019-12-31
Feature/s3 tileserver

kamicut:
This refactors a few things in order to move away from the OSMESA API service:
- Add a settings page and API for settings such as S3 keys/secret pairs
- Add a cache to read/write to
- Move the OSMESA service to read from the cache
- Refactor the user extent route to read from s3 instead of the OSMESA API


LanesGood:
@kamicut Are there default entries for these settings that would ship with a public scoreboard? And if so, does there need to be a "reset" or "restore to defaults" button?

LanesGood:
@kamicut I see that you have a commit stating "Remove references to OSMesa API" but this is still the label for each field in the settings page. Should these be more abstract? Or alternatively, should there be some explanatory text here that helps admins understand what this section is for, and how to use it? (I would think yes)

pages/admin/settings.js
```diff
@@ -0,0 +1,129 @@
...
+      await this.props.getAuthenticatedUser()
+      await this.props.getSettings()
+      this.setState({ loading: false })
+    } catch (err) {
+      console.error(err)
+      this.setState({ loading: false })
+    }
+  }
+
+  async handleSubmit (data) {
```
guidorice: Consider encrypting settings before saving to database, especially because AWS keys are involved. There is a Js crypto api that could be used https://caniuse.com/#search=crypto . The settings would of course need to be decrypted in memory to use the key+secret. `handleSubmit` may not be the place to do that, but I wanted to log it somewhere.

api/src/routes/settings.js
```diff
@@ -0,0 +1,74 @@
...
+    const { user } = req
+
+    // only admins are allowed to view settings
+    if (!user || !user.roles || !validateRole(user.roles, 'admin')) {
+      return res.boom.unauthorized('Not authorized')
+    }
+
+    // Get data from the cache first
+    let settingsMap = {}
+    let data = JSON.parse(cache.exportJson())
```
guidorice: Consider using cache.keys() and then iterate them and cache.get(...) to extract the settings instead of JSON export + parsing.  However, if this `get()` function is only called once, then it may not be worth refactoring and it would be fine to leave as is.

api/src/db/seeds/development/settings.js
```diff
@@ -0,0 +1,15 @@
+const { cache } = require('../../../config')
+
+const settings = {
+  'osmesa-db': 'postgres://postgres@localhost:5434/postgres'
+}
+
+exports.seed = (knex) => knex('settings').del() // delete entries
+  .then(async () => {
+    const promises = Object.keys(settings).map(key => {
+      cache.put(key, settings[key])
```
guidorice: Consider adding an expiration to all the uses of `cache.put()`. There needs to be some way to expire and/or invalidate the cache. The package's docs say

> If time isn't passed in, it is stored forever

kamicut: If we are using the cache only to keep the settings in memory, wouldn't invalidating the values require some new logic to pull the settings again from the database? In what cases do we need to invalidate the cache?

guidorice: @kamicut yes I see what you mean. The only use case I can think of is: what if the admin wanted to update a setting with a SQL update instead of using the UI admin interface? The in database and the in-memory settings would stay different until restart. If that is fine, then ignore this one.

pages/admin/settings.js
kamicut: This has been handled by the DB settings model encrypting and decrypting on storage.

api/src/routes/settings.js
kamicut: This has been rewritten

api/src/db/seeds/development/settings.js
kamicut: That makes sense, I think we want to make sure that "if you are hitting the database you know what you're doing" and that's not available to app admins but rather devops. In the case of devops I would lean on the side of "if you're touching the database you are probably restarting the service"

api/src/utils/cryptr.js
```diff
@@ -0,0 +1,33 @@
+// https://gist.github.com/vlucas/2bd40f62d20c1d49237a109d491974eb
+
+const crypto = require('crypto')
+
+class Cryptr {
```
guidorice: On a call with @kamicut we decided that this class would be a good candidate for a unit test (to make sure it's wrapping of the node crypto api is solid).

api/src/models/settings.js
```diff
@@ -0,0 +1,48 @@
...
+const { cryptr } = require('../config')
+const { identity } = require('ramda')
+
+let encrypt = cryptr.encrypt.bind(cryptr)
+let decrypt = cryptr.decrypt.bind(cryptr)
+
+// When testing, do not encrypt or decrypt settings
+//
+if (process.env.NODE_ENV === 'test') {
+  encrypt = decrypt = identity
```
guidorice: @kamicut I have been running in my dev env: `time yarn test` with this code block present and also commented out, and it does not seem to offer any performance boost to testing. It always takes about "real  1m9.549s" approximate length of time. Consider removing this conditional to simplify the code in settings.js?

As I think you mentioned on the call: this would also mean the settings.js model is tightly coupled to the Cryptr class. But the model is also coupled with the database connection so settings.test.js is already more of an integration test than a unit test. As a separate effort we could look at dependency injection as an alternative?

kamicut: @guidorice I like your suggestion for decoupling, but I feel that it might be a larger refactor since our models are all coupled to the database. We could take a look at our tests in another larger issue maybe?

guidorice: This sounds good to me.

LanesGood:
@kamicut to address my comment above, what do you think of the simple explanatory text in this screenshot? If agreed, I'll push. (Also updated the heading size)

image

kamicut:
@LanesGood I agree with the style, you could you push it up?

kamicut:
When debugging this on the tileserver, we noticed that the number of users dropped to 0, but stats were available for individual users. I realized that the `users_clock` was erroring because it could not read the OSMESA_DB setting from the database. 

On the server, the process starts and we do an initial cache load with the settings from the database, but we had not implemented that in the server. We can do an initial cache load for the `users_clock` process as well in the main run function of the file, unless there's a better way to initialize the cache for both of these processes @guidorice

kamicut:
Hey @guidorice could you take another look at the review since I've made some new commits? They address two issues found on staging:

- tiles were not showing up because of a typo: we were trying to deconstruct `req.user` instead of `req.params.user`
- the `users_clock` process was erroring when trying to connect to the osmesa database. This is because the way the cache is constructed it needs to be loaded when the process restarts. Maybe this is not very robust, so would like your thoughts on this

guidorice:
@kamicut is the latest deployed to staging? that would make testing easier.

api/src/users_clock.js
```diff
@@ -82,7 +84,13 @@ async function usersWorker () {
 
 // Run
 if (require.main === module) {
-  usersWorker()
+  dbSettings.list().then(settings =>
+    // load the cache
+    settings.forEach(({ setting, value }) => {
+      cache.put(setting, value)
+    })
+  )
```
guidorice: This code is duplicated, so I would consider refactoring it to a single function. But actually the more I consider it, it would be better if it did not have exist at all. I think this implementation is pointing to an antipattern within the memory cache usage. If there is a cache miss, the memory cache could hit the models/settings module to attempt to restore the cache from the database (lazy loading). Pre-loading a cache manually,  and having to do so in multiple  places seems error prone and easy to forget again.

kamicut: Yes I agree, I think that manually loading the cache is an anti-pattern. I'll think of a better way here for a cache miss. The issue that i've been facing is mostly related to order of dependencies loaded in the database and so manually loading from the database was the easiest fix.

api/src/tm_clock.js
```diff
@@ -50,7 +52,13 @@ async function tmWorker (isCmd) {
 
 // Run
 if (require.main === module) {
-  tmWorker(true)
+  dbSettings.list().then(settings =>
+    // load the cache
```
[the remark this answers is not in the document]
kamicut: I'll update the JSDoc; "clock process" is an artifact of running the service in development on heroku, where "clock process" is a type of process in a 12-factor app https://devcenter.heroku.com/articles/scheduled-jobs-custom-clock-processes#custom-clock-processes

kamicut:
@guidorice I could update system services and README in this PR, or would you rather revisit it all as part of a possible move to docker based deployment mechanism?

guidorice:
@kamicut yes, agreed we could  revisit the system services files and README outside of this PR.

api/src/users_clock.js
guidorice: @kamicut it appears memory-cache does not support lazy loading, however, it could be used wrapped in a closure like this:

https://gist.github.com/guidorice/adf0628871796be3324a54f3803725b9

guidorice:
@kamicut I approved- please do open a separate enhancement issue for the cache implementation.

Issue #471, opened 2020-01-09
Commit yarn.lock into git again


guidorice:
This should make builds more likely to be reproducible on CI and deployment. Is recommended  here:

https://yarnpkg.com/en/docs/yarn-lock

kamicut:
Closed by #472

PR #472, opened 2020-01-09
Add yarn.lock

kamicut:
Close #472


guidorice:
@kamicut I am curious why is circleci image version required to change in this PR? I am in favor of it, just was not sure if there was a specific reason related to the yarn.lock.

> Close #472

Did you mean Closes #471 ?

kamicut:
Hey @guidorice sorry just saw your message. I upgraded node to 12.x and that's why we need to change the circleci image.

guidorice:
@kamicut interesting it appears that adding yarn.lock back to git causes more secvulns to be tracked by github, via the yarn.lock files I guess:

https://github.com/developmentseed/scoreboard/network/alerts

Should we open a separate issue to track these upgrades?

PR #458, opened 2019-12-10
MapRoulette integration

guidorice:
Work in progress! Do not merge

Implements #457

Issue #457, opened 2019-12-09
MapRoulette + Scoreboard Integration


guidorice:
Integrate MapRoulette (MR) into Scoreboard and display a user’s active MapRoulette tasks on their dashboard.

Subtasks

- [ ] Decide on Semantics of what is a "campaign" on the frontend in MR terms; is it a challenge? a single task to complete?
- [ ] Perform Gap Analysis between Scoreboard database schema and the MR api. Are any columns or fields missing or mismatched?
- [ ] Add any database migrations to scoreboard, if any new columns or tables needed.
- [ ] Implement MR Task Manager classes (see tm_clock.js, tm.js for task manager abstraction)
- [ ] Update the user-facing UI to display MR data
- [ ] Update the admin UI to enable create/config of MR tasking manager

MapRoulette api docs
https://maproulette.org/docs/swagger-ui/index.html?url=/assets/swagger.json

guidorice:
###  Decide on Semantics of what is a "campaign" on the frontend in MR terms;  is it a challenge? a single task to complete?

#### Decision

After discussion and whiteboarding with @LanesGood , we decided that MapRoulette's **Challenge** entity is the appropriate level to source data from, into Scoreboard's Campaign entity.

#### Background

Scoreboard has, at a high level, the entities **Tasker**, **Campaign** and **Assignments**. Tasker is an abstraction for sourcing data from a concrete Tasking Manager instance.

Currently, Scoreboard can utilize Campaigns that originate from Projects in OpenStreetMap US Tasking Manager (tasks.openstreetmap.us). Some of the project data is cherry-picked and copied into the Scoreboard database, and also `tm_id` is used as a foreign key to the Task Manager's Project. 

example project tm_id
`Campaign <- tm_id (117) <- https://tasks.openstreetmap.us/project/117`

Tasker can be extended to support additional Tasking Manager instances, for example:

- HOT Tasking Manager
- MapRoulette 
- Other Tasking Managers...

MapRoulette has different terminology and different data model entities, so here are some concepts:

- MapRoulette **Project**: seems to be a grouping mechanism for a set of  challenges created by an admin or power user. Almost like a Folder. Existing Projects have names like "[username]'s project". Note this is different than the OSM Tasking Manager's concept of a Project.
[...]

PR #477, opened 2020-01-16
MapRoulette integration

guidorice:
Supersedes  #458 
Implements #457


guidorice:
I am closing this PR because it will be more convenient to use shared repository development model rather than working on a fork of scoreboard repository.

PR #493, opened 2020-01-31
Feature/add cli cmd

kamicut:
Adds a command to add random edits to a user

```
./api/bin/scoreboard create-edits --osm-id {your user id}
```


guidorice:
@kamicut is going to make an update to this PR involving the osmesa users table which is required for the scoreboard app to roll up the edits and statistics correctly.

PR #495, opened 2020-02-04
Upgrade knex.js to latest.

guidorice:
Fixes CVE-2019-10757

CVE-2019-10757
high severity
Vulnerable versions: < 0.19.5
Patched version: 0.19.5

knex.js versions before 0.19.5 are vulnerable to SQL Injection attack. Identifiers are escaped incorrectly as part of the MSSQL dialect, allowing attackers to craft a malicious query to the host DB.


guidorice:
This does not effect us because we are not using MSSQL, however this should quiet the warning coming from github.

@kamicut for your review.

Issue #476, opened 2020-01-14
Enhancements to README and .env.sample file


guidorice:
As a newcomer to Scoreboard development, here are some areas the README could be improved:

> Node 8.4

This should match current node requirements (#469 )

> Env file

The contents of `.env.sample` file is fewer lines than what's listed in the table in this section of the readme. Some lines of `.env.sample` are commented out. These should be made consistent with the readme. Consider also this section could explain what environment variables are needed to have a minimum working development environment.

> Postgres Setup, Test, Serve, Lint, etc.

These seem to be sub-sections of "Installation and Dev" section, so consider making them a smaller heading level: H3, H4, etc.

> ./api/bin/scoreboard create-admin-user --username {your osm username}

Last time I tried, this command fails unless you first  have signed in to the Scoreboard UI using your OSM credentials. Consider documenting that here, and/or create a bug issue. Do we want the `create-admin-user` to work around that edge case. Probably yes.

> Production

Maybe add 2 subsections: Manual Deployment, and Containerized Deployment? If there is any work done on Docker-compose or Kubernetes, perhaps we could link to it and say it's a work-in progress?

## Additions to consider

- Mention Next.js, give overview re: why it was chosen, and what features are currently being used?

guidorice:
In discussions with @kamicut we agreed it would be best to remove the environment variables table from the Readme, so they only have to be maintained in one place, and instead add liberal comments into the `.env.sample` organized into sections. Here are some notes, work in progress:

```
OSM_TEAMS_CLIENT_ID=teams-oauth-key  --> required to have yours
OSM_TEAMS_CLIENT_SECRET=teams-oauth-secret --> required to have yours
OSM_CONSUMER_KEY=your-oauth-key  --> required to have yours
OSM_CONSUMER_SECRET=your-oauth-secret  --> required to have yours

NODE_ENV=development --> default value
#TM_URL=https://tasking-manager-url --> this is now a setting in ui+db  (remove)
#TM_HASHTAG=tasking-manager-prefix --> this is now is a setting in ui+db  (remove)
#OSMESA_API=https://link-to-osmesa-api --->  this is now is a setting in ui+db (remove)
#APP_URL=http://localhost:8181 --> default value, can override

OSM_DOMAIN=https://www.openstreetmap.org -->default value, can override
SESSION_SECRET=super-secret-sessions -->default value, can override
DATABASE_URL=postgres://scoreboard:test@localhost:5433/scoreboard --> default value, can override
[...]

PR #509, opened 2020-02-20
Create Team Page

### LINKED ISSUE #504: Create team
Users who have signed in and integrated the Teams application via Connect Teams #501 are able to Create Team from the Teams index page.

Create Team enables the addition of all existing information of a team, including:
- Add team name, hashtag, description, location

kamicut:
* Closes #504 
* Adds a create team form page at `scoreboard/create-team`
* Creates a details form that can be reused by edit teams

### DIFF
--- a/components/teams/LocationInput.js
+++ b/components/teams/LocationInput.js
@@ -1,0 +1,22 @@
+import { useState } from 'react'
+import { Map, TileLayer, CircleMarker } from 'react-leaflet'
+
+export default function LocationInput ({ value, onChange }) {
+  const [ zoom, setZoom ] = useState(1)
+
+  return (
+    <Map style={{ height: '300px' }}
+      center={value}
+      zoom={zoom}
+      onViewportChanged={({ center, zoom }) => {
+        setZoom(zoom)
+        onChange(center)
+      }}>
+      <TileLayer
+        attribution='&amp;copy <a href="http://osm.org/copyright">OpenStreetMap</a> contributors'
+        url='https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png'
+      />
+      <CircleMarker center={value} color='blue' radius={20} />
+    </Map>
+  )
+}
--- a/components/teams/LocationInput.js
+++ b/components/teams/LocationInput.js
@@ -4,7 +4,7 @@ import { reverse } from 'ramda'
 
 export default function LocationInput ({ value, onChange }) {
   const [ zoom, setZoom ] = useState(1)
-  let newCenter = reverse(JSON.parse(value).coordinates)
+  const newCenter = reverse(JSON.parse(value).coordinates)
 
   return (
     <Map style={{ height: '300px' }}
@@ -12,7 +12,7 @@ export default function LocationInput ({ value, onChange }) {
       zoom={zoom}
       onViewportChanged={({ center, zoom }) => {
         setZoom(zoom)
-        let toGeojson = `{
+        const toGeojson = `{
           "type": "Point",
           "coordinates": [${center[1]},${center[0]}]
         }`
--- a/components/teams/LocationInput.js
+++ b/components/teams/LocationInput.js
@@ -1,22 +1,28 @@
 import { useState } from 'react'
 import { Map, TileLayer, CircleMarker } from 'react-leaflet'
+import { reverse } from 'ramda'
 
 export default function LocationInput ({ value, onChange }) {
   const [ zoom, setZoom ] = useState(1)
+  let newCenter = reverse(JSON.parse(value).coordinates)
 
   return (
     <Map style={{ height: '300px' }}
-      center={value}
+      center={newCenter}
       zoom={zoom}
       onViewportChanged={({ center, zoom }) => {
         setZoom(zoom)
-        onChange(center)
+        let toGeojson = `{
+          "type": "Point",
+          "coordinates": [${center[1]},${center[0]}]
+        }`
+        onChange(toGeojson)
       }}>
       <TileLayer
         attribution='&amp;copy <a href="http://osm.org/copyright">OpenStreetMap</a> contributors'
         url='https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png'
       />
-      <CircleMarker center={value} color='blue' radius={20} />
+      <CircleMarker center={newCenter} color='blue' radius={20} />
     </Map>
   )
 }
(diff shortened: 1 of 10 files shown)


guidorice:
@kamicut  I am attempting to connect Scoreboard w/ osm-teams running locally, but getting a bit confused about updating my `.env` file. I am noticing that the Scoreboard Readme, when compared the `.env.sample` file it differs quite a lot. (See my notes about the varying set of environment variables in #476) The Readme and the .env.sample should be 100% consistent when it comes to environment variables.

Also I think the Scoreboard Readme should get a new section describing what steps are needed to setup Scoreboard as an osm-teams oauth app. Specifically what is the Callback URL for Scoreboard? i.e. what to enter into placeholder `https://mayapp/callback` in the OAuth2 settings form in osm-teams.

cc @LanesGood

components/teams/LocationInput.js
```diff
@@ -0,0 +1,28 @@
+import { useState } from 'react'
+import { Map, TileLayer, CircleMarker } from 'react-leaflet'
+import { reverse } from 'ramda'
+
+export default function LocationInput ({ value, onChange }) {
+  const [ zoom, setZoom ] = useState(1)
+  let newCenter = reverse(JSON.parse(value).coordinates)
```
guidorice: Consider using `const` instead of `let`.

components/teams/LocationInput.js
```diff
@@ -0,0 +1,28 @@
...
+
+  return (
+    <Map style={{ height: '300px' }}
+      center={newCenter}
+      zoom={zoom}
+      onViewportChanged={({ center, zoom }) => {
+        setZoom(zoom)
+        let toGeojson = `{
```
guidorice: Consider using `const` instead of `let`.

components/teams/TeamDetailsForm.js
```diff
@@ -0,0 +1,31 @@
+import { useForm, Controller } from 'react-hook-form'
+import dynamic from 'next/dynamic'
+
+const LocationInput = dynamic(() => import('./LocationInput'), { ssr: false })
```
guidorice: I am seeing a parsing error from vscode here and also in CampaignCard.js

> Parsing error: Unexpected token import

Are you seeing the same @kamicut? It seems the syntax is correct for Next.js and ES2020, but we may need to update either `.eslintrc.json` or open an issue on devseed-standard.

components/teams/TeamDetailsForm.js
```diff
@@ -0,0 +1,31 @@
...
+
+export default function TeamDetailsForm ({ onSubmit, details }) {
```
guidorice: Consider adding a test for TeamDetailsForm in test-ui modules.

components/teams/LocationInput.js
```diff
@@ -0,0 +1,28 @@
+import { useState } from 'react'
+import { Map, TileLayer, CircleMarker } from 'react-leaflet'
+import { reverse } from 'ramda'
+
+export default function LocationInput ({ value, onChange }) {
```
guidorice: Consider adding a test for LocationInput in test-ui modules.

pages/create-team.js
```diff
@@ -0,0 +1,51 @@
+import React from 'react'
+import { connect } from 'unistore/react'
+import { actions } from '../lib/store'
+import TeamDetailsForm from '../components/teams/TeamDetailsForm'
+import Router from '../lib/router'
+
+function CreateTeam (props) {
```
guidorice: Consider adding a test for CreateTeam in test-ui modules.

pages/create-team.js
```diff
@@ -0,0 +1,51 @@
...
+  const handleSubmit = async (data) => {
+    try {
+      await props.createTeam(data)
+      Router.push('/teams')
+    } catch (e) {
+      console.error(e)
```
guidorice: Consider removing this try/catch block. Unhandled exceptions should get logged to console anyways.

At first it looks like this block could fail silently for the user if there was an exception. But I believe `createTeam` prop is coming from `lib/store/actions/teams.js`, which has state management/notification and logging of the exceptions in it's `createTeam()`. If that is indeed case, then the try/catch block here is not doing anything and could be removed entirely.

guidorice:
re: the issues with the Readme, the `.env.sample`, and getting  osm-teams running locally, @kamicut got me all setup and running. I am going to put that knowledge into #476 which is assigned to me.

components/teams/TeamDetailsForm.js
kamicut: @guidorice I'm not using a different eslint, are you getting an error when you run `yarn lint` ?

pages/create-team.js
kamicut: I think the idea here was to make sure that a Router push doesn't happen when there is an error (copied pattern from another part of the codebase). I'm going to take a look at the store and make sure that an exception doesn't still trigger the Router code.

guidorice: Line 11 should not be reached at all if there is an exception from line 10- regardless of the try/catch block being there.

components/teams/TeamDetailsForm.js
guidorice: @kamicut ah good point to check: `yarn lint` reports no errors, but in my vscode editor, it's showing errors like:

Screen Shot 2020-02-26 at 2 26 07 PM

Apparently vscode think's it's running eslint but doing it differently than the devseed-standard script. Are you not seeing the same problem?

kamicut:
@guidorice I've added a test for CreateTeam that loads all the children component except the map (because it requires extensive mocking) and makes sure they can be mount onto a page. I think writing tests for the sub components would have to have us rethink our entire testing approach for all components since we only test containers right now.

pages/create-team.js
guidorice: @kamicut can you respond or edit on this one?

guidorice:
@kamicut re: component unit testing, I agree completely about mainly just testing containers. I think what you did in the new test 'Create Team renders without crashing' is sufficient. 

In my experience, even if we were using [Enzyme](https://enzymejs.github.io/enzyme/) for React tests it still only works very will with shallow mounted components, and deeply nested components are much more difficult.

Incidentally, `enzyme` is in in package.json but I don't see it being imported anywhere. Should we remove that dependency?

pages/create-team.js
kamicut: I removed the try catch @guidorice

Issue #198, opened 2018-12-18
Single Team page


giblet22:
Building off of #197 aggregate stats at a team level similar to how it's done for campaigns

ex.  Team A has mapped X buildings, Y coastlines, Z km of roads, etc

and be able to sort team members by edits

LanesGood:
Currently proposed Teams table, borrowing from the Campaigns table:

| Rank | User | Campaigns | Roads(Km) | Points of Interest | Buildings | Railways(Km) | Coastlines(Km) | Waterways(Km) | Changesets | Edits |   |
|------|------|-----------|-----------|--------------------|-----------|--------------|----------------|---------------|------------|-------|---|

After conversations with @kamicut I understand teams statistics to be untied to teams - that is, we can't determine Km of Coastlines for a user's activity related to a team campaign, as there's no way to mark on a changeset that it should only be counted to a particular team.

However, I wonder if we can sort "Rank" within a team to see the highest contributors?

I also wonder if Campaigns is a useful statistic. Ideally, this would be the campaigns worked on by a user for that particular team. Unfortunately, I think this is also not possible at the moment, unless campaigns could be tagged with team names, and then teams read from those tags. 

@kamicut does the above sound correct to you?

LanesGood:
@kamicut I've created the mockups below as two possible options. I think that while it could liven up the page, the user avatar introduces a risk of monotony given that very few users have a profile image. Both options incorporate a full team stats table, while v1 also includes some elements of the individual user dashboard.

v1
![TeamStats_v1](https://user-images.githubusercontent.com/12634024/69754714-9ab0e300-1124-11ea-94e1-6357678270ee.png)

v2
![TeamStats_v2](https://user-images.githubusercontent.com/12634024/69754715-9ab0e300-1124-11ea-8d49-79d18ba28ce8.png)

What do you think?

kamicut:
I think this looks pretty good! I do prefer smaller avatar images. However, we could always create generated avatars to fill in missing profile images (we can detect this in the user profile request). In any case I think we can start work on the user tables for statistics and then figure out how to display the members.

LanesGood:
@kamicut I've added back the "Team Scoreboard Panel" section into v2, posted below as v3. I'm aware that the "Active Campaigns" section may also not be possible, but we could change this to campaigns.

One additional idea: just one table with Members, rank, country, and total # of edits. Clicking an "Expand stats" button would then expose the feature counts for each member.

[...]

PR #519, opened 2020-03-03
Add team statistics to teams route

guidorice:
Relates to #198,  #527 

- Adds an `osmesaStats` data structure to the response in route `/teams:id`
- Does not update `api.yml` because that is a larger task covered by #527 .

### DIFF
--- a/api/src/services/osmesa.js
+++ b/api/src/services/osmesa.js
@@ -1,7 +1,15 @@
+const pg = require('pg')
+// conversion of postgresql bigint (int8) to JS number (from https://github.com/knex/knex/issues/387)
+pg.types.setTypeParser(pg.types.builtins.INT8, 'text', parseInt)
 const knex = require('knex')
 const AWS = require('aws-sdk')
-
-const { find, propEq, reduce, assoc } = require('ramda')
+const {
+  assoc,
+  find,
+  propEq,
+  reduce,
+  sum
+} = require('ramda')
 
 const { cache } = require('../config')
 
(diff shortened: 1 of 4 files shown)


guidorice:
@kamicut I think this is ready to review. I rebased it to clean up noise commits, so make sure to delete your local branch if you had already looked at it locally.

I did not change the campaigns query in `api/src/routes/teams.js` - and I am unsure about that campaigns query- if it is correct as is, or how/when the `team_id` gets populated in the `team_assignments` table. And if it's the same team id as osm-teams? 

We should pair on this. It might be a blocker,  and also I want to look for any other missing stats which might be required by #198 . Because I think the "# of campaigns for a team", and  "active campaigns for a team" might need additional work.

But I still think this PR could proceed anyways but it is up to you.

api/src/services/osmesa.js
```diff
@@ -1,7 +1,15 @@
+const pg = require('pg')
+// conversion of postgresql bigint (int8) to JS number (from https://github.com/knex/knex/issues/387)
+pg.types.setTypeParser(pg.types.builtins.INT8, 'text', parseInt)
```
kamicut: Was there a test failure that caused this? Do we need this anywhere else in the app as a setting?

guidorice: @kamicut great question! There was no test failure, but in the new `getTeamStats` method, it sums over the user statistics, so needs to parse strings into integers for some columns. This is one possible workaround, which seemed cleanest to me. I got it from the linked github issue in the knex repo.

### Background

This workaround is needed because any PostgreSQL  bigint column will be returned by Knex as a string, because the bigint values could be greater than JavaScript `Number.MAX_VALUE`. 

I think a safe bet is we are not going to have values larger than 2<sup>1024</sup> coming from this particular query. If we think it's possible for that to happen, then it is likely the whole app, not just this module, will need to be refactored to use a math library that supports large numbers, like the new BigInt feature, or some other existing math library.

> Do we need this anywhere else in the app as a setting?

The greater question is how are we dealing with very large numbers in the app overall? It's possible in other places in the app that bigint queries are being returned and strings and are being handled differently, like just displayed as raw string values. 

So one option is making a bigger scoped issue doing the due diligence about handling large numbers coming from PostgreSQL. 

But I am also comfortable just scoping this change to this module, and not automatically  doing it everywhere knex is querying PostgreSQL. I assume this  `setTypeParser` call only effects this module, because it did not cause any test regressions elsewhere in the app. But I do not actually know that 100% certainty.

guidorice:
> This looks good to me. Out of curiosity why did you pull the functions outside the class?

It was not strictly necessary, purely stylistic :) I got carried away when adding JsDocs to the existing code, and making one of the functions re-usable.  Then I realized that there were several functions on the class that were internal implementation details, not part of the public API of the class. 

So I hoisted them up to be hidden in the closure of the Js module. There are a few  alternatives and one would be to prefix them with `_blah` underscore names, but that is not as clean as hiding them in the closure.

guidorice:
@kamicut do you have any feedback about this question?

> I did not change the campaigns query in api/src/routes/teams.js - and I am unsure about that campaigns query- if it is correct as is, or how/when the team_id gets populated in the team_assignments table. And if it's the same team id as osm-teams?

> We should pair on this. It might be a blocker, and also I want to look for any other missing stats which might be required by #198 . Because I think the "# of campaigns for a team", and "active campaigns for a team" might need additional work.

api/src/services/osmesa.js
kamicut: I don't think we have values anywhere near that resolution unless users start mapping the entire world at the picometre (10^-12) scale 😂. It might be however useful to centralize string to number conversions somewhere. Let's discuss this on a call.

guidorice:
> "# of campaigns for a team", and "active campaigns for a team"

We discussed in the scrum meeting and decided to defer on that question and go ahead and merge this PR.

PR #523, opened 2020-03-11
Feature/users country sort

LanesGood:
Initial progress on users table country sort

To do:
- [x] check stub tests for accuracy
- [x] test filter in action with user data

### DIFF
--- a/api/tests/api/users.test.js
+++ b/api/tests/api/users.test.js
@@ -135,6 +135,34 @@ test('Sort users alphabetically z-a', async (t) => {
   t.deepEqual(resCopy, names)
 })
 
+test('Sort users by country alphabetically a-z', async (t) => {
+  const response = await request(app)
+    .get('/scoreboard/api/users/stats/?q=&page=1&sortType=Countries A-Z&active=false')
+    .expect(200)
+  const users = await db('users')
+    .select('country')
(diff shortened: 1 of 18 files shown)


LanesGood:
@kamicut this is ready for review. Can you check that the tests in `api/tests/api/users.test.js` are correct?

api/tests/api/users.test.js
```diff
@@ -135,6 +135,34 @@ test('Sort users alphabetically z-a', async (t) => {
...
+})
+
+test('Sort users by country alphabetically z-a', async (t) => {
+  const response = await request(app)
+    .get('/scoreboard/api/users/stats/?q=&page=1&sortType=Countries Z-A&active=false')
+    .expect(200)
+
+  const users = await db('users')
+    .select('country')
+    .orderBy('country', 'desc')
```
kamicut: In this and the previous test, you have to make it orderByRaw so it matches

LanesGood: @kamicut thanks, this has been updated

LanesGood:
@kamicut I've updated this to enable the requested featue of a non-sortable Users table. In doing so I uncovered that our nonSortable prop wasn't affecting the tables as intended. This is now working as expected for Users + Countries tables. Campaign tables are still being sorted as expected.

PR #525, opened 2020-03-12
Update centos/deploy.sh

### LINKED ISSUE #524: deploy.sh missing build step
scoreboard/deployment/centos/deploy.sh should have an `npm build` step.

guidorice:
Based on the last centos deployment, which we did manually, here are
some improvements:

- add missing `npm run build` step
- add variable and echo prompt for which git tag to use
- use `bash -e` so script will stop if something fails

Other than that, this deploy script looks fine and we should use it on
next deployment.

Closes #524

### DIFF
--- a/deployment/centos/deploy.sh
+++ b/deployment/centos/deploy.sh
@@ -10,26 +10,32 @@
 #
 
 GIT_TAG=${1:-master}
+NGINX_PATH=/var/www
 
 echo "using git tag or branch: ${GIT_TAG}"
 echo
 
 cd ~
+
+# find the tag or branch
 rm -rf scoreboard
 git clone https://github.com/developmentseed/scoreboard.git
-cd scoreboard
+pushd scoreboard
 git checkout ${GIT_TAG}
 
-# installation
+# install & build steps
 yarn install
 yarn run migrate
 yarn run build
+popd
 
-# move to nginx folder
-cd ../..
-suffix=$(date +%s)
-sudo mv /var/www/scoreboard /var/www/scoreboard_$suffix || echo 'this is the first deployment'
-sudo mv scoreboard /var/www/
+# update nginx directory
+datestamp=$(date +%F)
+if [ -d "${NGINX_PATH}/scoreboard" ]; then
+    # backup previous deployment
+    sudo mv -v ${NGINX_PATH}/scoreboard ${NGINX_PATH}/scoreboard_${datestamp}
+fi
+sudo mv -v scoreboard ${NGINX_PATH}
 
 # restart services
 sudo systemctl restart scoreboard-api
--- a/deployment/centos/deploy.sh
+++ b/deployment/centos/deploy.sh
@@ -13,9 +13,9 @@ cd scoreboard
 git checkout ${GIT_TAG}
 
 # installation
-npm install
-npm run migrate
-npm run build
+yarn install
+yarn run migrate
+yarn run build
 
 # move to nginx folder
 cd ../..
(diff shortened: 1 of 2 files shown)


deployment/centos/deploy.sh
```diff
@@ -1,14 +1,21 @@
-#!/bin/sh
+#!/bin/bash -e
+
+# change this to the current semver tag, e.g. v1.8.0
+GIT_TAG=master
```
kamicut: Should we turn this into a shell argument?

kamicut:
Should we delete the repo and start from scratch at every deploy?

deployment/centos/deploy.sh
```diff
@@ -1,14 +1,21 @@
...
 
 cd ~
 rm -rf scoreboard
 git clone https://github.com/developmentseed/scoreboard.git
 cd scoreboard
-git checkout master
+git checkout ${GIT_TAG}
 
 # installation
 npm install
```
kamicut: Should we use yarn?

deployment/centos/deploy.sh
guidorice: Yes, I thought of that too, so thanks for calling it out 👍  will do.

deployment/centos/deploy.sh
guidorice: @kamicut good catch. Because we have `yarn.lock` in the repository, we won't get it's benefits by using `npm` to install javascript dependencies.

A side effect is yarn will need to be added to the deployment docs I am working on. 

will make these changes 👍

guidorice:
> Should we delete the repo and start from scratch at every deploy?

@kamicut I guess you mean these steps?

```bash
rm -rf scoreboard
git clone https://github.com/developmentseed/scoreboard.git
```

I left those in the script because it's just a sanity check. I have seen web projects where crufty old files get leftover on the filesystem for whatever reason (usually someone not using version control , or they are build artifacts or something old). And starting from a new git clone prevents that.

kamicut:
yeah, I want to make sure that there are no cache or `.tmp` that the database or next are using that could speed up new builds. Let's double check that the deploy using this script is idempotent

guidorice:
@kamicut let's aim for correct and idempotent builds. build speed should not matter.

guidorice:
@kamicut I addressed those change requests and also did some additional cleanup.

deployment/centos/deploy.sh
```diff
@@ -1,27 +1,41 @@
...
+# update nginx directory
+datestamp=$(date +%F)
+if [ -d "${NGINX_PATH}/scoreboard" ]; then
+    # backup previous deployment
+    sudo mv -v ${NGINX_PATH}/scoreboard ${NGINX_PATH}/scoreboard_${datestamp}
+fi
+sudo mv -v scoreboard ${NGINX_PATH}
 
 # restart services
 sudo systemctl restart scoreboard-api
```
kamicut: Maybe we restart scoreboard-timer.service for good measure

guidorice: @kamicut good point. I will look into this- either it is required or not, not for good measure :)
I think that `scoreboard-timer.service` is not running, rather the `scoreboard-timer.timer` starts it periodically but will double-check this.

guidorice: @kamicut actually I believe `sudo systemctl restart scoreboard-timer.service` and `sudo systemctl restart scoreboard-timer` are identical commands. I updated the syntax though, see my last commit.

PR #526, opened 2020-03-16
[Draft] Template team page

LanesGood:
This PR introduces some additional components and styles according to the team mock. The teams.js page does not yet work given the need to correctly connect data from the API.

Target UI:

TeamPage_Connected

**Edit - updated mockup above 3/24/20**

PR #530, opened 2020-03-23
Add OSM-teams statistics page

### LINKED ISSUE #198: Single Team page

guidorice:
Add OSM-teams statistics page at `/teams/{id}`

Implements (partial) #198
Relates to #526

### Notes to reviewers

- The `Blurb` component works a little differently than in other places in the app. (zero valued metrics will be omitted from the english phrase).

- Some manual setup is required to be able to see anything on the Team stats page:

1. Create a team via osm-teams UI (or via Scoreboard- should work either way).

1. Via the osm-teams UI assign some osm ids to your team. You should use ids that are in the seed data in scoreboard+osmesa. To get a list of useful osm ids, query the osmesa db like:
    ` psql -h localhost -U postgres -p 5434 --command "select id, name from users"`

1. Associate your team with a campaign in the scoreboard db. The UI for this may be broken so instead use SQL to do it. First get a list of valid campaign ids to choose from:
    ` PGPASSWORD=test psql -h localhost -U scoreboard  -p 5433 scoreboard --command "select id, name from campaigns"`
    then insert a record into `team_assignments`. For example for team id 1 and campaign id 114:
    ```PGPASSWORD=test psql -h localhost -U scoreboard  -p 5433 scoreboard --command "INSERT INTO public.team_assignments (team_id, campaign_id, created_at, updated_at, team_priority) VALUES (1, 114, '2020-03-31 15:32:40.873000', '2020-03-31 15:32:43.880000', 1);"```
1. The Last Refreshed value will display without some additional inserts (at bottom)

1. Now if you visit `/teams/id`, e.g. `/teams/1` you should see the team stats and team campaigns.

### Known issues

- [x] CSS Styling is not finished
- [x] Join Team button action not implemented ... enhancement deferred to #539 
- [x] Link from Campaign name to the Campaign details page seems broken
- [x] Assigned date (may) be wrong on the Team Campaigns table.
- [x] Last Refreshed does not always display a value. Possibly related to #534 

### Last Refreshed dates

 This data should be seeded in the dev environment but it is not.

 ```bash

psql -h localhost -U postgres -p 5434 --command "INSERT INTO public.refreshments (mat_view, updated_at) VALUES ('user_statistics', '2020-03-05 00:00:00.000000');"
psql -h localhost -U postgres -p 5434 --command "INSERT INTO public.refreshments (mat_view, updated_at) VALUES ('hashtag_statistics', '2020-03-01 00:00:00.000000');"
psql -h localhost -U postgres -p 5434 --command "INSERT INTO public.refreshments (mat_view, updated_at) VALUES ('country_statistics', '2020-03-02 00:00:00.000000');"
psql -h localhost -U postgres -p 5434 --command "INSERT INTO public.refreshments (mat_view, updated_at) VALUES ('hashtag_user_statistics', '2020-04-01 00:00:00.000000');"

PGPASSWORD=test psql -h localhost -U scoreboard  -p 5433 scoreboard --command INSERT INTO public.user_update (id, last_update) VALUES (1, '2020-04-01 00:00:00.000000');

```

### DIFF
--- a/components/teams/TeamBlurb.js
+++ b/components/teams/TeamBlurb.js
@@ -62,7 +62,6 @@ export default function Blurb ({
     )
   }
   return <h2 className='header--small width--shortened list--block'>
- 
     { teamName } has mapped{' '}
     { fragmentList.map((fragment, i) => {
       return i === fragmentList.length - 1
--- a/components/teams/TeamBlurb.js
+++ b/components/teams/TeamBlurb.js
@@ -2,6 +2,24 @@ import React from 'react'
 import Link from '../../components/Link'
 import { formatDecimal, formatKm } from '../../lib/utils/format'
 
+/**
+ * Display a blurb about team statistics. Only displays non-zero metrics,
+ * so it can form various sentences like:
+ *
+ * "Since 2015, Team America has mapped 13,218 km of roads."
+ * "Since 2018, merica has mapped 2000.0 km of roads, 12,312 buildings, 945
+ * points of interest, and 23423.0 km of waterways."
+ *
+ * @param buildingsMappedCount
+ * @param editTimes
+ * @param firstYearEdited
+ * @param poiCountMappedCount
+ * @param roadsKmMapped
+ * @param teamName
+ * @param waterwaysKmMapped
+ * @returns {React.PureComponent}
+ * @constructor
+ */
 export default function Blurb ({
   buildingsMappedCount,
   editTimes,
@@ -9,8 +27,15 @@ export default function Blurb ({
   poiCountMappedCount,
   roadsKmMapped,
   teamName,
-  waterwaysKmMapped
+  waterwaysKmMapped,
+  coastlinesKmMapped
 }) {
+  if (!firstYearEdited) {
+    return <h2 className='header--small width--shortened list--block'>
+      { teamName } has not made any mapping edits yet. Explore{' '}
+      <Link href='/campaigns'>active campaigns</Link> to get started!
+    </h2>
+  }
   const fragmentList = []
   if (roadsKmMapped) {
     fragmentList.push(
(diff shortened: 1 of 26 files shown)

### CODE REVIEW

**Comment by guidorice** [comment]
This is definitely not finished yet, but I could use some feedback

@LanesGood about the styling, definitely could use your help
@kamicut general architecture and code review

**Comment by LanesGood** [comment]
@guidorice I've updated the styles on this page. The `Campaigns` class is where the relevant styles live, so I've kept that; the styles in `teams.scss` were used for the teams index page, specifically for the connect banner. I can rename these two files.

File: components/teams/TeamBlurb.js [hunk 1]
```diff
@@ -0,0 +1,73 @@
...
+        <><mark>{ formatKm(waterwaysKmMapped) }</mark> of waterways</>
+    )
+  }
+  if (coastlinesKmMapped) {
+    fragmentList.push(
+        <><mark>{ formatKm(coastlinesKmMapped) }</mark> of coastlines</>
+    )
+  }
+  return <h2 className='header--small width--shortened list--block'>
+    Since <mark>{firstYearEdited}</mark>,{' '}
```
LanesGood, line 65:
```suggestion
 
```

@guidorice I suggest that we remove the `FirstYearEdited` prop from the blurb, as this reflects the earliest team member individual edit, rather than anything to do with the team itself. We should still keep the prop, as its needed to show whether a team has any edits at all.

guidorice: @LanesGood great catch- will make this change.

pages/team.js
```diff
@@ -1,70 +1,125 @@
...
+      created_at: teamCreated,
+      lastRefreshed,
+      osmesaStats,
+      campaigns
+    } = team
+    const { buildingsMappedCount, poiCountMappedCount, roadsKmMapped,
+      waterWaysKmMapped, coastlinesKmMapped } = calculateBlurbStats(osmesaStats.teamStats)
+    const allMemberEditTimes = flatten(osmesaStats.memberStats.map(({ edit_times }) => edit_times))
+    const firstYearEdited = getYear(head(allMemberEditTimes.map(t => t.day).sort(compareAsc)))
+    const exportDataFilename = this.slugger.slug(teamName || 'team export', false)
```
kamicut, line 36: What's the rationale behind using `github-slugger`?

guidorice: @kamicut just personal preference for having filenames that are not full of spaces and special characters. Would be happy to remove it if you do not like it.

kamicut: It was just out of curiosity

PR #547, opened 2020-04-09
Team Stats page: empty state, edge cases

### LINKED ISSUE #542: Team Stats page needs empty state
In #530 a first pass at a Team Stats page was implemented. One thing that was left out was an "empty state":

- What are the scenarios where the Team Member stats list is empty? Is the user shown prompts for what to do? Add users to team? What if team has users, but there are no map edits?

- Are there scenarios where the Team Campaigns list is empty? Is the user shown prompts for what to do? Manually associate Team with Campaign through the Admin settings?

Relates to #198 

### remaining tasks (edited, from comments on this issue)

- [x]  If there are team members, but have no map edits, Blurb links to /campaigns to explore.
- [x]  ~If there are NO team members, then link to the Edit Team route for adding members.~ this is actually not possible because a team will always have >= 1 member (moderator is required).
- [x] ~implement Join Team button for non-editors~ # this is tracked in #539 
- [x] implement Edit Team button for editors
- [x] Team Member Stats table shows all members regardless of whether osmesa has mapping statistics for them.
- [x] ~If a team is private and a user isn't authenticated OR if a user isn't a member, they shouldn't be able to see that team's page at all.~ # is blocked by https://github.com/developmentseed/osm-teams/issues/47 and we are not implementing this for the current milestone.
- [x] If a team does not have campaigns and a user is not the moderator, the empty state should read "{TEAM NAME}'s moderator hasn't assigned any campaigns" or "There are no current assignments for this team."
- [x] If a team does not have campaigns, the empty state should link to the Edit Team route, `/teams/:id/edit-details` and say "There are no current assignments for this team. Add campaign assignments here {link}"

### LINKED ISSUE #545: Team Stats page Edit Team link, conditional Team campaigns section
Building on #530, and copied from #198:

- For logged-in users who 1) have connected the Teams app, and 2) are members of the team:
Team campaigns

- For logged-in users who 1) have connected the Teams app, and 2) are moderators of the team:
Edit team link

guidorice:
Closes #542 
Closes #545

### DIFF
--- a/api/src/services/teams.js
+++ b/api/src/services/teams.js
@@ -8,8 +8,13 @@ const { OSM_TEAMS_SERVICE } = require('../config')
  * Methods to grab data from OSM Teams
  */
 class OSMTeams {
+  /**
+   * Constructor
+   *
+   * @param osmid {number} openstreetmap user id
+   */
   constructor (osmid) {
-    this.osmid = osmid
+    this.osmid = Number(osmid)
   }
 
   /**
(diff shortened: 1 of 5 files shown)


api/src/services/teams.js
```diff
@@ -83,6 +88,45 @@ class OSMTeams {
...
+   * @param id {number} The osm-teams team id
+   * @returns {Promise<boolean>}
+   * @async
+   */
+  async canEditTeam (id) {
+    try {
+      const [orgOpts, teamOpts] = await Promise.all([
+        this.addAuthorization({
+          method: 'GET',
+          uri: `${OSM_TEAMS_SERVICE}/api/organizations/${id}`,
```
kamicut, line 105: I think you mean for this `id` to be the environment variable `OSM_TEAMS_ORG_ID` and not the team id, but you should merge in the latest from develop into your branch so you can access that.

guidorice: 👍 good catch

PR #559, opened 2020-04-19
Add page for team moderation

kamicut:
This adds a new page at `teams/:teamId/edit-moderators` that allows the current moderator to modify the moderator list. I added it as a new page because it is interactive rather than being done through a form. We should link to it from either the team edit page, or from the single team page.

This PR is based on `feature/edit-team`, so if we merge that in first we should change the PR target.

### DIFF
--- a/api/src/routes/index.js
+++ b/api/src/routes/index.js
@@ -52,6 +52,8 @@ router.post('/teams', teams.post)
 router.get('/teams/:id', teams.get)
 router.put('/teams/:id', teams.put)
 router.delete('/teams/:id', teams.del)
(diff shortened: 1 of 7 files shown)


kamicut:
Known issue:
- Double clicking on add or remove moderator can pop a "failed to update team" toast because of it trying to hit the database twice. This can be solved by disabling the button after click.

guidorice:
@kamicut would you open an enhancement issue for this?

>  This can be solved by disabling the button after click.

It's not really unique to this PR. We should solve this in the same way for all submit / action buttons in the app.

PR #562, opened 2020-04-22
Teams index page: Add public teams and filters for user's teams

### LINKED ISSUE #503: Add a user's member and moderated teams to Teams page when signed in + connected
When a user has !) signed into Scoreboard and 2) connected the Teams application (via #501) the main Teams index page should contain three tables: All Teams (public), `My Teams`, and `Teams Moderating`

TeamsIndex_A0

guidorice:
### Notes to reviewer
Here are some UI states to please test:

- Unauthenticated
- Authenticated
- There are teams you are a moderator of
-  ... teams you are a member of
- ... teams you are not a member of

Closes #503 

### known issues

- the table’s column sort order is not preserved when filters are changed 
- client side filtering won’t scale for thousands of teams. but perhaps can be converted to be a database filter like the campaigns listing, when we have thousands of teams to deal with.

### DIFF
--- a/api/src/index.js
+++ b/api/src/index.js
@@ -43,25 +43,15 @@ swaggerDocument.info.version = pckg.version
  */
 
 let sessionConfig = {
+  name: 'scoreboard',
   secret: SESSION_SECRET,
   resave: false,
-  saveUninitialized: true,
-  expires: new Date(Date.now() + (30 * 86400 * 1000))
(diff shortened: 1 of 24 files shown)


guidorice:
> This happens with both toggle boxes, they don't restore state.

@kamicut good catch- I'm on this now.

PR #566, opened 2020-04-27
Propagate loggedIn prop to teams page

kamicut:
This propagates the loggedIn prop that is created in `_app.js` for the navbar and header to all pages. Ideally, we would also do this for the authenticated user props and attach it to all page props but that's a larger refactor. 

I've also fixed static links in the banner, and an error that can happen in the state when the user is logged in but there is still no token.

### DIFF
--- a/api/src/routes/teams.js
+++ b/api/src/routes/teams.js
@@ -17,9 +17,16 @@ async function list (req, res) {
   try {
     const { user: { id: osmId = null } = {} } = req
     const teamService = new OSMTeams(osmId)
-    const teams = JSON.parse(await teamService.getTeams())
-    const canCreate = await teamService.canCreateTeam()
-    return res.send({ teams, canCreate })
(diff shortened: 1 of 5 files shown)


api/src/routes/teams.js
```diff
@@ -17,9 +17,16 @@ async function list (req, res) {
...
-    return res.send({ teams, canCreate })
+    const teams = await teamService.getTeams()
+    let canCreate = false
+    try {
+      canCreate = await teamService.canCreateTeam()
+    } catch (e) {
+      if (e.message !== 'No token for user') {
+        console.error(e)
+      }
+    }
```
guidorice, line 28: Consider adding a comment here to explain why the `no token for user` situation is being suppressed and we are instead sending a result of `canCreate: false`.

kamicut:
@guidorice do you mean the Teams page? or the Teams service?

Issue #527, opened 2020-03-16
Swagger api/docs/api.yml is outdated


guidorice:
There are many routes in `api/src/routes/index.js` which are not documented in `api/docs/api.yml`.

guidorice:
Also `api/README.md` duplicates much of the swagger api docs, so the readme should be replaced with instructions for viewing the stagger api docs.

cc @LanesGood @necoline @jvntf

guidorice:
Oops, rather, this script outputs to the README.md

```json
    "docs": "swagger-markdown -i api/docs/api.yml -o api/README.md",
```

PR #587, opened 2020-08-05
openapi spec for timeseries api

guidorice:
- Update swagger-ui component to @latest.
- Remove badly formatted and often outdated `api/README.md`. This is a manual step and it seems redundant with having swagger-ui in the app already.

@LanesGood if you pull down this branch, and do the usual setup, plus a `yarn install` then you should see the updated swagger-ui, at http://localhost:8181/api/docs,  and  there will see a preview of the spec for the timeseries endpoint.

Relates to #527 
Relates to #603 

Alternatively you can run swagger-ui by adding to the docker-compose.yml

```yaml
  openapi:
    image: swaggerapi/swagger-ui
    environment:
      - SWAGGER_JSON=/docs/api.yml
    volumes:
      - type: bind
        source: ./api/docs
        target: /docs
    ports:
      - 8888:8080
```

cc @kamicut @jvntf @necoline


necoline:
@guidorice Do you think there is value in keeping a `api/README` that provides an overview of how the api is structured as well as instructions for finding and using `http://localhost:8181/api/docs`?

LanesGood:
@guidorice this is looking good. If no users or hashtags are passed as filter, would this request all data? I'm guessing we would want to limit this for performance reasons if so. Also to confirm, countries is not possible here because of lack of geometry, right?

guidorice:
@LanesGood really good points! per our phone call just now: we'll add the business logic for it's required to have at least one filter applied (so entire db is not returned) and also, I believe it should be possible to add a countries filter, because osmesa has a `changeset_countries` details table. I will add that to the openapi spec. 

cc @necoline @jvntf

guidorice:
> @guidorice Do you think there is value in keeping a `api/README` that provides an overview of how the api is structured as well as instructions for finding and using `http://localhost:8181/api/docs`?

@necoline yes, I will add in a generic readme, explaining how to discover the openapi spec at least. thanks

guidorice:
@necoline for your review. After this is merged, I will work on implementing the timeseries api endpoint in a new branch.

Please look at ` /api/timeseries` in the api.yml, in particular the allowed filters, and the response data structure.

necoline (approved):
The docs look very good! Thanks for all the additional cleanup. 
I'm not sure I understand the implications of the bin interval option.

guidorice:
@necoline thanks for review.  Going ahead to merge this. However, we should have a call to talk about the binInternal option in more detail. In  there is no binning interval, then you are not dealing with time series data, you just have a resultset with start and end date. There is a difference. Maybe something was missed from the UI mockups about the time series aspects of this? 🤔

necoline:
ahh, yep yep ok when I was reading this, my head was stuck in the table display, and the filtering options but I get what you are saying here. I think there are a few UI elements that still need to be ironed out, but we can discuss this elsewhere.

PR #589, opened 2020-08-06
MapRoulette Task Manager and Data

jvntf:


api/src/services/tm_types/map_roulette.js
```diff
@@ -0,0 +1,133 @@
...
+    })
+  }
+
+  /*
+   * For consistency, use Projects to refer to Scoreboard projects
+   * MapRoulette uses Projects :: Challenges
+   * Challenges are treated as Campaigns.
+   * TODO this currently statically returns the first 50 objs
+  */
+  async getProjects () {
```
necoline, line 40: It looks like there are a few campaigns/challenges coming through without any features. 
I only noticed as I was paginating through the `mr` campaigns in the UI. They show up around the 4th page.

jvntf: hm strange. looks like on first load of the page, no geometry shows up in the map but if you click on a campaign, the geometry appears. but when I look at the response, all the entries have geometry.

jvntf: On the client side, the geometry is getting delivered on first load. so looks like there is some bug preventing it from display on the initial view.

LanesGood:
@jvntf hyped to see this get going. I don't think this is a blocker, but we'll need to update the `seed` database as well. This might involve updating osm-fabricator - I'm unsure. At the moment, I see the following when running `yarn seed`:

```
~/github/scoreboard $ yarn seed
yarn run v1.22.4
$ knex seed:run --cwd api/src/db
Working directory changed to ~/Github/scoreboard/api/src/db
Using environment: development
Challenge 52 does not have available metadata
Challenge 202 does not have available metadata
Challenge 200 does not have available metadata
Challenge 213 does not have available metadata
``` 

As for the MapRoulette challenges themselves, its great to see this data get pulled into Scoreboard. I see that the current data model doesn't fit the MR challenges. I can work on a new mock of the ideal MR challenge page, which we can work on in a different branch. Can you retitle this PR to reflect that this is MapRoulette data + Tasking Manager, but not the full integration?

api/src/services/tm_types/map_roulette.js
necoline: Maybe this is better addressed in a different PR?

LanesGood:
@jvntf additionally, do we know what stat is being read for the Scoreboard "Mapped %" value? Many of these seem to be at 100%, but I'm not sure if this is accurate. We'll also need to convert or change the "Validated" stat to an appropriate alternative.

jvntf:
> @jvntf hyped to see this get going. I don't think this is a blocker, but we'll need to update the `seed` database as well. This might involve updating osm-fabricator - I'm unsure. At the moment, I see the following when running `yarn seed`:
Can you clarify please? As far as I understand the `seed` database, it was pulling down tasks from OSM Task Manager (tasks.openstreetmaps.org). This PR adds pulling down from Map Roulette as well. Is there something missing?

The 'Challenge X does not have metadata' message is because some of the Map Roulette challenges don't have available info on percentage mapped and what not. 



> @jvntf additionally, do we know what stat is being read for the Scoreboard "Mapped %" value? Many of these seem to be at 100%, but I'm not sure if this is accurate. We'll also need to convert or change the "Validated" stat to an appropriate alternative.

This value is from this MapRoulette [endpoint](https://maproulette.org/api/v2/data/challenge/%7Bid%7D) which has no documentation :/ 
Example respose
```
[{"id":85,"name":"Texas Cemeteries","actions":{"total":6019,"available":0,"fixed":3645,"falsePositive":1242,"skipped":0,"deleted":0,"alreadyFixed":1129,"tooHard":3,"answered":0,"validated":0,"disabled":0,"avgTimeSpent":98153,"tasksWithTime":2}}]
```

guidorice:
@jvntf I have not looked at the code very closely, but I can see you are making solid progress on this.  🚀 Hopefully my docs and earlier PRs made a reasonable starting point, however feeble they were.  👍

PR #590, opened 2020-08-07
Downgrade nodemon

jvntf:
I am confused about this but today upon running the server I found that the most current version of `nodemon` is incompatible with some other packages. Downgrading it to `1.18`.


jvntf:
closing this because i think there is a different issue

PR #593, opened 2020-08-14
Maproulette campaign data

jvntf:
This PR makes attempts at a few things. This is WIP, comments / feedback really needed and appreciated here before I move on.

Some of this has been discussed in slack, will reiterate for clarity.

The goal is to integrate MapRoulette data into scoreboard.
To do this I added a MR class in #589 which loads MR campaigns on `yarn seed`

This PR adds 
- Loading some MapRoulette users on `yarn seed`
     - this may be completely unnecessary? im getting closer, but still rather unclear on what data is expected to be available in the deployed env. But this PR pulls in users from MR just as POC, and then tries to extract statistics from OSMesa (fails most of the time because the users are not there).
- Adds conditional method of getting Campaign data
    - Prior to this, Scoreboard assumes campaign data from all TMs will be accessible in OsMesa (via `osmesa.getCampaign()`) and it is resolved with a `hashtag` and users are found via the same `hashtag`
     - MapRoulette does not seem to be able to be queryable with the same logic
     - So instead this PR adds a TM class function to get project stats from the Map Roulette API (these are different than OSM stats). These stats are somewhat in line with the mockups for this page posted by @LanesGood -> list of users and their generalized levels of contribution for a campaign.
    - I am not particularly happy about some of this approach which takes a `if (map roulette) {} -> else {}` route, once we agree on a general solution

I looked into updated `osm-fabricator` to include MR data, but have held off because the kind of per-statistics / members of the osm schema do not seem to be available from MR. If I have this wrong please lmk!!

### DIFF
--- a/components/common/Table.js
+++ b/components/common/Table.js
@@ -133,6 +133,7 @@ export default function Table (props) {
   if (sortable && props.initialSortColumn) {
     useEffect(() => toggleSortBy(props.initialSortColumn, 'descending'), [])
   }
+  console.log(headers)
 
   return (
     <table>
@@ -154,7 +155,7 @@ export default function Table (props) {
                 </th>)
                 : (
                   <th key={column.id} {...column.getHeaderProps()}>
-                    <div className={column.header === 'number' ? 'table-align-right' : ''}>{column.Header}</div>
+                    <a className={column.Cell.name === 'formattedNum' ? 'table-align-right' : ''}>{column.Header}</a>
                   </th>)
             )
           }
--- a/components/common/Table.js
+++ b/components/common/Table.js
@@ -133,7 +133,6 @@ export default function Table (props) {
   if (sortable && props.initialSortColumn) {
     useEffect(() => toggleSortBy(props.initialSortColumn, 'descending'), [])
   }
-  console.log(headers)
 
   return (
     <table>
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ docs-build
 node_modules
 node_modules.nosync
 .idea
+api/tests/tapes
--- a/api/src/db/seeds/development/settings.js
+++ b/api/src/db/seeds/development/settings.js
@@ -2,6 +2,7 @@ const { cache } = require('../../../config')
 const { put } = require('../../../models/settings')
 
 const settings = {
+  // default
   'osmesa-db': 'postgres://postgres@localhost:5434/postgres'
 }
 
--- a/api/src/db/seeds/development/users.js
+++ b/api/src/db/seeds/development/users.js
@@ -1,4 +1,7 @@
 const usersClock = require('../../../users_clock')
+const limit = require('p-limit')(100)
+const rp = require('request-promise-native')
+const { merge } = require('ramda')
 
 const { cache } = require('../../../config')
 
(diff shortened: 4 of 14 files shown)


jvntf:
Refactored as per [issue comment](https://github.com/developmentseed/scoreboard/issues/457#issuecomment-675508950)

Todo:

- [ ] move schemas to backend
- [ ] conditionally render summaries

jvntf:
@LanesGood  ready for review

jvntf:
thanks @LanesGood 
I've made these changes bar the last one (checking if user exists before linking). Do you mean we should only include a users stats in the list if they exist in the database? Or just disable the link if they don't? I think in the production environment they should always exist, but i do ofc see your point

LanesGood:
@jvntf updates look good. For users, I think only make the username a link if the user exists in the database - but yes, still include user in table regardless. 

A few more questions/change requests:
- Do we know if Avg Time Spent is working? None of the test challenges seem to have a value here
- Do we get any data on task validation? If not, we should probably change the "Validated" line on the campaign card on /campaigns
- Can we right-align the headings of the single-row table? I'm responsible for the code on this - but I can't quite figure out why the `table-align-right` class is not showing up
- Let's remove the tooltips from all columns in `maprouletteChallengeSchema` - I don't think that these need to be included

components/common/Table.js
```diff
@@ -133,6 +133,7 @@ export default function Table (props) {
   if (sortable && props.initialSortColumn) {
     useEffect(() => toggleSortBy(props.initialSortColumn, 'descending'), [])
   }
+  console.log(headers)
```
necoline, line 136:
```suggestion
```

jvntf:
With [this](https://github.com/developmentseed/osm-fabricator/pull/2) PR on OSM Fabricator, we can get some test data loaded for maproulette challenges, with user stats. There are some quirks I ran into, would appreciate some input from team.

I edited and built OSM Fabricator as `developmentseed/fabricator` and tagged as `latest`, running on port `5432:5432` as specified in the Fabricator README.
Before running n `docker-compose up -d` in score, I changed the `osmesa` container to run on port `5432:5432` to match the local version. This results in an error `Port 5432 has already been allocated`. 
As a solution, I allowed Docker to build scoreboard's `osmesa` using port `5434:5442`, and then changed the DB url in [this settings file](https://github.com/developmentseed/scoreboard/blob/develop/api/src/db/seeds/development/settings.js) to look at port `5432` and therefore use my modified Fabricator instance.

Finally I had to manually refresh materialized views: 
```REFRESH MATERIALIZED VIEW hashtag_user_statistics
REFRESH MATERIALIZED VIEW user_statistics
``` etc inorder to get user stats for maproulette challenges to show up.

sorry for this in the weeds post! let me know your thoughts 
cc @guidorice

**Comment by guidorice** [comment]
@jvntf sorry you had to do that hack of settings.js -- that should definitely not be necessary, and I'd like to come up with some steps for testing that that doesn't involve patching that settings module.. I think the problem is the scoreboard and osm-fabricator documentation is bad, and there is also some confusion about docker containers vs images. I will post more here once I can make sense of it.

**Review by guidorice — CHANGES_REQUESTED** [review]
@jvntf using your changes to osm-fabricator, from https://github.com/developmentseed/osm-fabricator/pull/2 ,  I made a `pg_dump` of that  osm-fabricator database, and tried running scoreboard against it for it's osmesa service. I ran into an issue which is in scoreboard running `yarn seed` throws a sql exception:

```
Error while executing "/Users/alex/repos/scoreboard/api/src/db/seeds/development/users.js" seed: insert into "users" ...
error: duplicate key value violates unique constraint "users_osm_id_unique"
```

We should make sure all the steps listed in the scoreboard readme still work OK.  Can you look into this? Let me know what you think. 

I am going to open up a new scoreboard PR which simplifies the handling of the mocked osmesa database, so this problem should be easier to reproduce/test there.

see also https://github.com/developmentseed/osm-fabricator/blob/master/build.sh which is something we missed yesterday, to update the .sql script in osm-fabricator. After we get this fixed up, we can (optionally) patch that in osm-fabricator.

**Comment by jvntf** [comment]
@guidorice i wasnt able to reproduce this. I also did not have this error when i tested #601. 
I ran
```
docker-compose down
docker-compose prune
rm -rf .tmp/db
docker-compose up
yarn migrate
yarn seed
```
This error indicates that there is some userName with more than one osm_id present in the join of MR and Task Manager data. We can protect against this in the osm fabricator scripts and update the pg dump used to create the new `init` files?

guidorice:
>  But this PR pulls in users from MR just as POC, and then tries to extract statistics from OSMesa (fails most of the time because the users are not there).

@jvntf this situation will be much improved in osmesa going forward once their users table foreign key bug is resolved.  In addition in the deployment environment, scoreboard downloads a list of known users at startup (populate_users.js). I hope that what you are experiencing  is just scoreboard's rough development environment. Continuing to review this today...

guidorice:
Testing w/ #601 

> Ran 5 seed files
> ✨  Done in 103.50s.

🎉

guidorice (approved):
@jvntf nicely done! The campaigns page looks and works great. I clicked through on a MR task to a user's page, and got a user not found error. However as was noted- this is probably an issue with the dev environment. After merging to develop we can do some testing on staging. I will take care of merging the other PR #601 .

PR #595, opened 2020-08-14
Fix wrong http error code when s3 fetch of .mvt fails.

guidorice:
- res.boom.serverError() was not a function, so use the s3 statuscode and the boom library to
return a more relevant api response.

- upgrade express-boom to get the boomify() function.

- wrap the osmesa.tiles() call with some additional error handling.

Before review:

- [x] test on staging environment, by changing the s3 credentials temporarily, because in scoreboard local development the s3 client is not even used at all.

### DIFF
--- a/api/src/routes/extents.js
+++ b/api/src/routes/extents.js
@@ -1,24 +1,42 @@
 const osmesa = require('../services/osmesa')
 
 /**
+ * For error handling the response cannot have the vector tiles response headers
+ *
+ * @param res
+ */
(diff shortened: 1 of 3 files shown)


jvntf:
@guidorice How should I test this?

guidorice:
@jvntf there is no test coverage for this functionality, and it's actually has conditional logic to run some other code in the development environment- which is pretty weird.... so I can talk you through testing it on the staging server.

PR #598, opened 2020-08-28
Add/org permissions

necoline:
## Completed:
- new routes for organization management
- new org management page
- user search 
- ability to add managers
- table to display owners and managers and change or remove status

## Known issues:
- ~I'm running into auth issues again with the teams api, I had it all working in the past, I'm not sure what changed~
  - update this was caused by my silly decision to test the functionality of the req by demoting myself to manager and not owner 🤦 After I forcibly added myself back, everything was fine
- ~button alignment on team page~
- ~nav button on left-hand menu in `/organization` I'm actually not sure where this is meant to link to~

## Testing instructions
- locally run osm teams
- set `OSM_TEAMS_SERVICE` to local instance of osm teams
- run scoreboard and log in
- navigate to `/teams`
- create a team if you don't have one already
- on `/teams`, navigate to `Org Management`
- test out user search and status changes

## Mockups:

image

image


guidorice:
@necoline I am running into some trouble testing this. I see the new UI at http://localhost:8181/organizations but when I try to add a new manager via the search box, I get an error dialog which flashes briefly, then a page reload. Over in my node console, I see

```
 body: {
      statusCode: 400,
      error: 'Bad Request',
      message: 'insert into "organization_owner" ("organization_id", "osm_id") values ($1, $2) - duplicate key value violates unique constraint "organization_owner_organization_id_osm_id_unique"'
    }
```

where the request was to `PUT /api/organizations/23/addOwner/10499924 HTTP/1.1`

10499924 is my osm user id for guidorice. I was already an owner of this org. So it seems like it's trying to re-insert the owner record each time the Save Changes button is submitted?

guidorice:
> duplicate key value violates unique constraint "organization_owner_organization_id_osm_id_unique"

We decided the best solution is to improve the osm-teams api to support PUT method better: https://github.com/developmentseed/osm-teams/issues/187

jvntf:
My only suggestion is w/r/t the issue that @necoline mentions above -> issues arise when a user demotes themselves. Should we prevent this altogether? Should a user be allowed to demote themselves? / or remove themselves / or remove themselves as the last member of an org? Should we require orgs to have at least 1 owner?

jvntf:
also wondering if this `reset` button should come above the `Create New Team` button rathe than below 
image

guidorice:
> or remove themselves as the last member of an org? Should we require orgs to have at least 1 owner?

The osm-teams api enforces,  IIRC, there must be at least one owner and at least one moderator of each team.
This should probably be surfaced in the UI to prevent those interactions.

cc @jvntf @necoline

necoline:
Great points. I can make changes so that the option to remove the last owner is not available. The question of whether or not one should be able to demote themselves is a good one as well. I think the answer is yes, but it has consequences - being that you cannot re-promote yourself to owner. Only owners can add owners - this should also have some sort of visible error handling.

So my to-do list is as follows:
- [ ] remove option to `remove owner` if user is the last owner
- [ ] add a warning if the user is about to demote themselves from owner

_anything else?_

guidorice:
Problems I have run into testing this branch locally:

1. Org Management -> Add organization manager: 
  - Typed "foo" to filter down to user Natfoot, clicked "Add Manager".
  - Click Save Changes
  - Three error Toasts appear too quickly to get a screen capture.  Something about network error when attempting to fetch. (the page fully reloads making the javascript console and the error toasts difficult to capture). I recommend removing whatever code is causing a full page reload after an error toast, because it makes things extremely difficult to debug. I believe this is a longstanding issue though re: the error handling.

2. When I navigate back to the Org Management  button to navigate to http://localhost:8181/organizations
  - React.js error re: `TypeError: checkForSingleOwner is not a function`

> The above error occurred in the <ManageOrg> component:
>     in ManageOrg (created by c)
>     in c (created by Scoreboard)
>     in div (created by Layout)
>     in Layout (created by Context.Consumer)
>     in WithAlert(Layout) (created by c)
>     in c (created by Scoreboard)
>     in Provider (created by Scoreboard)
>     in r (created by Scoreboard)
>     in Container (created by Scoreboard)
>     in Scoreboard (created by AppWithRedux)
>     in AppWithRedux

   *I think I have a fix for this React error, I will post in a change request in the PR.*

### About my setup

- scoreboard branch `add/org-permissions` at 7c8ac22da505cb32915e930022dfd88a0087a180
- osm-teams branch `master` at c458b833ca21236f6568440f3694ea06e772d259 is also running locally
- scoreboard .env file has `OSM_TEAMS_ORG_ID=1`
- my osm-teams database has only one team and one org. I am 10499924 and I am set to be the owner.


```
osm-teams=# select * from organization
organization                organization_manager        organization_owner          organization_team          
organization_id_seq         organization_manager_id_seq organization_owner_id_seq   organization_team_id_seq   
osm-teams=# select * from organization;
 id | name | description |          created_at           |          updated_at           
----+------+-------------+-------------------------------+-------------------------------
  1 | nome | nome dev    | 2020-09-22 11:33:39.447651-06 | 2020-09-22 11:33:39.447651-06
(1 row)

osm-teams=# select * from organization_team;
 id | team_id | organization_id 
----+---------+-----------------
  1 |       1 |               1
(1 row)

osm-teams=# select * from team;
 id |     name      |    hashtag    |    bio     | privacy | require_join_request |          created_at           |          updated_at           |                      location                      | editing_policy 
----+---------------+---------------+------------+---------+----------------------+-------------------------------+-------------------------------+----------------------------------------------------+----------------
  1 | guido-mappers | mappin-nappin | hell world | public  | f                    | 2020-09-22 12:54:16.638293-06 | 2020-09-22 12:54:16.638293-06 | 0101000020E610000001000000103F5AC018BBB23087CD4340 | 
(1 row)

osm-teams=# select * from organization_owner;
 id | organization_id |  osm_id  
----+-----------------+----------
  1 |               1 | 10499924
(1 row)

osm-teams=# select * from organization_manager;
 id | organization_id | osm_id 
----+-----------------+--------
(0 rows)
```

cc @necoline @jvntf @LanesGood

pages/organizations.js
```diff
@@ -0,0 +1,223 @@
...
+      let membersWithRoles = users.map(user => {
+        return organization.organization.owners.find(owner => owner.osm_id === user.osm_id) ? (
+          { ...user, role: 'owner' }
+        ) : (
+          { ...user, role: 'manager' }
+        )
+      })
+      setMembers(checkForSingleOwner(membersWithRoles))
+    }
+  }, [users])
```
guidorice, line 50: This useEffect also depends on the `organization` variable from the closure. It may be stale unless you list it as well in the dependency like: `[users, organization]`.

pages/organizations.js
```diff
@@ -0,0 +1,223 @@
...
+
+  if (loading) {
+    return <LoadingState />
+  }
+  if (!authenticatedUser.loggedIn) {
+    return <NotLoggedIn />
+  }
```
guidorice, line 57: If these returns run, then const definitions from below will not be in the function component's scope, including `checkForSingleOwner()`.

guidorice: Consider moving the early-out logic as high as possible in the function component.

pages/organizations.js
```diff
@@ -0,0 +1,223 @@
...
+
+  /**
+ * Adds a new parameter to owner if they are the only owner in the list.
+ * @param {array} members - object containing all data listed on the user table.
+ */
+  const checkForSingleOwner = (members) => {
+    const owners = members.filter(member => member.role === 'owner')
+    const singleOwner = owners.length === 1 ? owners.shift() : null
+    return singleOwner ? members.map(member => member.osm_id === singleOwner.osm_id ? { ...member, singleOwner: true } : member) : members
+  }
```
guidorice, line 73: Alternatively, move `const checkForSingleOwner`before the 1st use of it in `useEffect()`.

pages/organizations.js
```diff
@@ -0,0 +1,223 @@
...
+          break
+        case 'manager':
+          if (findInitialRole('owners', member)) {
+            props.removeOwner(member.osm_id)
+          }
+          props.addManager(member.osm_id)
+          break
+        case 'none':
+          props.removeManager(member.osm_id)
+          props.removeOwner(member.osm_id)
```
guidorice, line 119: I am not sure if this matters at all, but my IDE is flagging warnings here re: `Promise returned from xxxx is ignored`
Screen Shot 2020-10-15 at 2 26 49 PM

guidorice (changes requested):

pages/organizations.js
guidorice: (probably unrelated to the observed problem though)

pages/organizations.js
necoline: Could this be a result of the requests that don't execute because of duplicates?

necoline:
I'm unable to reproduce the error toasts on `Save Changes` I've done the following to identify where this might be happening
- slowed down my internet to see if the flash of toasts was happening before I could see it
- placed debuggers on each promise catch
- investigated the network tab to identify request errors

guidorice:
@LanesGood @jvntf passing the torch over to you to review this PR. I think testing it on the staging server would be a good next step. Note that would also involve updating osm-teams on the staging server to have the latest api fixes.

LanesGood (commented):
@necoline I've spent some time trying to test this, and still didn't get past local osm-teams issues: https://github.com/developmentseed/osm-teams/issues/193

LanesGood:
@jvntf Can you attempt testing this locally? I'm not sure how feasible it is to test on staging given the existing teams org setup.

jvntf:
Tried testing this and get blocked by osm teams failing to build 
image
```
docker-compose version 1.27.4, build 40524192
Docker version 19.03.13, build 4484c46d9d
```

cc @necoline @kamicut

LanesGood:
@jvntf is this related? https://github.com/developmentseed/osm-teams/pull/194

jvntf:
@LanesGood not sure, i can try to build teams again off that hydra pr

LanesGood:
@kamicut this could use your review - testing either locally or on staging environment if possible.

kamicut:
I've tested this locally. On Chrome, this works perfectly. I do get the Network Error on Firefox, which is an `NS_BINDING_ABORT`, meaning that the page has refreshed before completing the request (to the API). Next I'm going to test this as a build; this might be the deployment server refreshing the page too early.

kamicut (approved):
There's a weird error on Firefox that doesn't allow "save changes" to go through, I think because the page refreshes and cancels the request. I will however approve this and we can investigate further in another ticket @necoline.

PR #604, opened 2020-09-17
Parse options when loading campaign

jvntf:


guidorice:
@jvntf can you add a bit of description to this PR? Some info like is it a bug fix, or new feature, or why this change was needed? Linking to existing issues or bug reports too is helpful. Thanks much

jvntf:
@guidorice yes, sorry about that.

This is a bug fix. Previously the `campaign` route would always use the public task manager type specific url to load campaign data. This fix adds in the `opts` such that a proxy url will override if it exists.

guidorice (approved):
@jvntf thanks 🎉

PR #607, opened 2020-09-23
Fix regression bug

jvntf:
Fix for

image


necoline:
I was looking into this bug at the same time as you and found it rather difficult to navigate without a clear description of all the props these component expect to receive. I'm not totally clear on what should be required and what should be optional, but if you think these additions make sense, go ahead and merge

jvntf:
@necoline added some docs

necoline (approved):
looks great! thanks for the addition

PR #609, opened 2020-09-23
[wip] Bump knex.js to latest.

guidorice:
[knex.js](https://knexjs.org/) has some bug fixes and new minor version. There was one breaking change re: `stream.close` which I searched for in the code and did not see any instances of.


guidorice:
💥

Issue #608, opened 2020-09-23
regression with yarn test?


guidorice:
When running scoreboard locally,  `yarn test` fails with errors like:

`- duplicate key value violates unique constraint "users_osm_id_unique"',`

This issue is to investigate why this is and see if it is a regression bug, or how we can get `yarn test` to run locally again.

Other notes:

- Strangely, the CI process still succeeds in running `yarn test`, e.g. https://app.circleci.com/pipelines/github/developmentseed/scoreboard/137/workflows/3cf8b86d-d3c6-4563-82e2-aa04fdd4180e/jobs/1928

- It seems possible this is related to the `yarn seed` and the work done in #458 but I have not been able to confirm that.

guidorice:
@necoline @jvntf let's all look at this as time allows- seems like kind of a blocker.

guidorice:
Fixed by #611

PR #611, opened 2020-09-24
Fix/608 reg tests

jvntf:
This makes some fixes against #608 but there are a few problems

The unique_osm_id issue has been fixed, by filtering out duplicate osm values. Now CI test results match local ones, still don't know why this wasn't the case earlier on.

This revealed other tests that are failing. I put the blame on [users_clock](https://github.com/developmentseed/scoreboard/blob/6feba21fa4e87a5966d2f254d6a2054cf3d9a874/api/src/users_clock.js). There is something strange happening with the `usersWorker` function and its multiple `await`s that results in a timeout (?) or some reason for a premature exit. Originally this function was using `pLimit` to cap it at `100` concurrent promises. I think this is a major component of the problem, something is incorrect and when we use a limit, promise values seem to be getting lost and thus the timeout. let me know if more explanation is needed.

cc @necoline @guidorice


jvntf:
it truly seems that adding print statements causes this to work, and without them the promises just get lost

guidorice:
> it truly seems that adding print statements causes this to work, and without them the promises just get lost

Hmmm 🤔 that is pretty weird. Let me  look at this when I get a chance. @jvntf  @necoline

guidorice:
In standup today we decided @jvntf is going to make this PR just specifically fix the issue in #608 re: duplicate key value violates unique constraint "users_osm_id_unique"'. This will make this PR easier to reason about and the commit history will be cleaner.

If there are additional failing tests, we think that that's already a problem existing in `develop` branch, so we will open up additional issues/PRs to fix those after this PR merged.

cc @necoline

jvntf:
This PR has been reverted to only address issue #608, caused by non-unique osm IDs. Fixing this reveals an additional problem causing tests not to run due to premature exiting, which will be addressed in #616 cc @guidorice @necoline

PR #613, opened 2020-09-26
Fix for token refresh

kamicut:
I updated `simple-oauth2` to a new version, which fixes a bug in how refresh tokens are saved [ref issue](https://github.com/lelylan/simple-oauth2/pull/328). This should solve our token refresh woes as it uses a new constructor interface for storing the refresh token.

We should still test this against old tokens on the staging site, and against our current hydra/osm-teams version. We might benefit from also updating hydra on the osm-teams side.

### DIFF
--- a/api/src/models/teams-access-tokens.js
+++ b/api/src/models/teams-access-tokens.js
@@ -12,13 +12,15 @@ function getToken (id) {
 }
 
 async function storeToken (tokenObject) {
-  const { access_token, refresh_token, expires_at, id_token } = tokenObject
+  const { access_token, refresh_token, expires_at, id_token } = tokenObject.token
 
   const { sub } = jwt.decode(id_token)
   let existingToken = await getToken(sub)
   if (existingToken.length > 0) {
     await db('teams_access_tokens').where('osm_id', sub).update({
-      access_token, refresh_token, expires_at
+      access_token,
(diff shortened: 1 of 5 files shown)


guidorice:
@kamicut would you go ahead to do the hydra update in osm-teams, and do some more testing with this branch? Then we can merge to develop and setup some time to test on the staging environment.

guidorice:
@kamicut would you add some comments to the PR about steps for testing? I think we need to change the hydra settings to expire tokens quickly, on the order of a few minutes, to enable testing of the code flow for refresh and saving.

Also can we write down a rollout strategy, because IIRC there are expired tokens in the scoreboard db in `teams_access_tokens`, and not sure if they will get refreshed, or they are "stuck" and we have to delete those records?
cc @LanesGood

kamicut:
This works on staging after updating hydra to v1.4! We can test it with a few accounts and then merge it in.

LanesGood:
@kamicut what do you need for final testing here, and who should test?

I attempted to connect my teams in the staging environment, but unfortunately receive the same error I highlighted in https://github.com/developmentseed/osm-teams/issues/193

LanesGood:
@kamicut can this be merged?

LanesGood (approved):

Issue #602, opened 2020-09-09
osm-fabricator needs to be removed from .circleci/config.yml


guidorice:
In #601 osm-fabricator was removed from the docker-compose.yml, but I noticed it's still being built and used in the circleci config, which should be cleaned up.

guidorice:
Closing in favor of #615

PR #601, opened 2020-09-04
Decouple from osm-fabricator.

guidorice:
After a some discussion w/ @jvntf, osm-fabricator was adding a bit of overhead and friction, so this removes that dependency, and replaces it instead with a postgres db which loads from an init.sql file (which is itself a
pg_dump of an osm-fabricator db).

This PR makes it all under the control of docker-compose, and so it should be clear what is going on just by looking at the scoreboard repo.

### Features

- osmesa service uses postgres 10 w/ no postgis (note: this exactly matches deployed osmesa db version)
- add to readme some  general info about osm-fabricator.
- enhance docker-compose.yml with a wait-for-ready script.
- `init.sql` this corresponds to the osm-fabricator image on dockerhub tagged `developmentseed/fabricator:latest`
- `init-with-mr.sql` this corresponds to @jvntf PR on osm-fabricator to add MR data https://github.com/developmentseed/osm-fabricator/pull/2

Therefore to test #593 it "should" be a matter of 😅 

1. merge in this branch into @jvntf branch
2. replace` init.sql` with `init-with-mr.sql`
3. `docker-compose down`
4. `docker container prune`
5. `rm -rf .tmp/pg`
6. `docker-compose up`
7. `yarn migrate, yarn seed...` test app, etc

Some quirks to note that I used in dumping from osm-fabricator, which uses postgis whereas this PR does not use postgis:

1. `pg_dump postgres://postgres@localhost/postgres --schema=public --if-exists --clean > scripts/osmesa-mock/init.sql`
2. Edited the .sql and remove the empty table "spatial" reference-- it for some reason pollutes the public schema.

### Next Steps

- [x] test with @jvntf branch for #593
- [x] after #593 is merged, replace init.sql with init-with-mr.sql and merge this.

Closes #599

PR #617, opened 2020-09-29
Migrate to Github Actions CI 

### LINKED ISSUE #615: migrate to github actions CI ASAP
This issue supersedes the osm-fabricator issue,  #602 

Because fabricator is not used in local development & testing anymore, and we are in a bad state now because tests seem to run differently in CICD versus local development (assuming it's because of the current `.circle/config.yml`, but not 100% sure).

guidorice:
This PR

1. Migrates from CircleCI to Github Actions CI
2. Removes osm-fabricator from the workflow (it is not used in the development environment anymore, and we want CI to match the dev environment closely as possible)
3. Runs lint, build and test on both node 12 (lts) and node 14 (current)

closes #615 
relates to #601 
relates to #602


jvntf:
I've replaced the `node` script `ci-schema.js` running via github actions with a `psql` command. i believe the problem with loading the `pg_dump` via `node pg` is something to do with `psql`'s ability to handle large files? i read some posts about `pg` failing to load dumps for some reason, not too familiar with that.

This change succeeds the `Setup osmesa db`, using the `init.sql` dump file.

A couple of the `campaigns` api tests were failing (namely, sort alphabetically). I think this is just a bug in the test scripts which use a `alphabeticalDiff` function but don't take case in to account. Addressed this by lower case-ing all the campaign names before sorting them. Note that the tests use a random subset (or seeds?).

I believe at this point the CI behaves the same as local dev environment. Tests are timing out, and 'failing', I imagine this is solved by #616.

jvntf:
I think this PR is _done_, after doing some tests in #618. We have a couple of problems that are sort of tangent to this PR.

First is the test time out issue addressed in #616. Tests fail in this PR because the tests timeout while the `before` scripts run to prepare the db for tests. For some reason, adding some arbitrary print statements alleviates this. I tested this in #618 but haven't merged that code here.

Second is w/r/t a non-deterministic bug in one in the [campaigns](https://github.com/developmentseed/scoreboard/blob/f6270041d539ffd632dcf69d67feb4beca33ef97/api/tests/api/campaigns.test.js#L136) test. I believe knex `orderBy`, which is used [here](https://github.com/developmentseed/scoreboard/blob/f6270041d539ffd632dcf69d67feb4beca33ef97/api/src/routes/campaigns.js#L78) is treating capital characters differently than `alphabeticalDiff` used in the test, s.t. this test sometimes fails. It seems to fail more consistently on CI rather than local (truly no idea why that is). 

I propose we approve or further solve #616, merge that here, merge this, and then address the alphabetical bug in a separate PR.

guidorice:
> I've replaced the node script ci-schema.js running via github actions with a psql command

@jvntf I like this approach! I had mistakenly assumed the `psql` client was not available on the CI runner.

guidorice:
@jvntf for your review. thanks for your assist on this!

@sethvincent if you have the powers, would you please remove the circle ci check (Stop Building) via the link 
https://circleci.com/gh/developmentseed/scoreboard/2050?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

jvntf (approved):

sethvincent:
@guidorice @jvntf circleci is removed. i deleted the circleci webhook, which should be all we need, but let me know if it keeps trying to run the check.

PR #622, opened 2020-10-14
Fix table formatting issue

jvntf:


LanesGood (changes requested):
@jvntf this looks great! I can confirm the table headers are correctly aligned.

I do see a "Failed to get campaign data" error for every TM campaign, and see an osmesa error in the terminal. This is unrelated to this PR, but is it related to bringing in maproulette campaigns? Error:
```
OSMesa Stats do not exist for this hashtag connect ECONNREFUSED 127.0.0.1:5434
Campaign 50-117, Failed to create response Cannot read property 'tasker_id' of undefined
OSMesa Stats do not exist for this hashtag Cannot read property 'campaign_hashtag' of undefined
TypeError: Cannot read property 'type' of undefined
    at module.exports (/Users/lane/Github/scoreboard/api/src/routes/campaign.js:103:86)
```

In Docker, I see
```
wait-for-db-ready_1 | Service osmesa:5432 did not start within 300 seconds. Aborting...
scoreboard_osmesa_1 exited with code 1
```
and 
```
osmesa_1             | 2020-10-14 17:27:21.759 UTC [1] DETAIL:  The data directory was initialized by PostgreSQL version 11, which is not compatible with this version 10.14 (Debian 10.14-1.pgdg90+1).
```

The container I'm connected to is `scoreboard_db_1` - has this been deprecated in favor of `wait-for-db-ready-1`? Do I need to upgrade the container's version of postgres to run the osmessa db? (And if so, how would I do that?)

Finally, now that we have multiple tables, I'd like to add table titles to distinguish what data is included. Can we go ahead and do that in this PR?

LanesGood (approved):
@jvntf thanks for getting me up and running with the local osmesa db - everything now looking and working well!

For the table titles, these should be "Challenge Task Progress" and "Challenge Leaderboard", respectively.

PR #623, opened 2020-10-14
Add paging loop for maproulette challenges

jvntf:
Addressing the issue of only pulling in 50 of the oldest maproulette challenges. Unfortunately i don't see a way to determine how many challenges exist through the API, other than requesting until we don't get results. The problem with doing this on seed is that there seem to be roughly 1mil challenges which takes quite a while to download and parse. I don't think it is a good idea to pre-fetch all of these vs maproullette, which requires the user to 'see more' before fetching the next page of results.

As a solution, I have set the scripts to fetch in descending order, sorted by popularity (get most relevant challenges) and increased the cap to 500.
Thoughts?

### DIFF
--- a/api/src/services/tm_types/map_roulette.js
+++ b/api/src/services/tm_types/map_roulette.js
@@ -62,7 +62,7 @@ class MapRouletteAPI {
       pe: true,
       ce: true,
       order: 'DESC',
-      sort: 'popularity'
+      sort: 'created'
     }
 
     if (this.opts.search_params) {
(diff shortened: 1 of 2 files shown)


LanesGood (changes requested):
@jvntf I played with the `extendedFind` parameters to get challenges sorted by created date in descending order. Is this a better solution so that we see the most recent challenges? I believe that's the sort order for Tasking Manager campaigns imported into scoreboard.

request url:
`https://maproulette.org/api/v2/challenges/extendedFind?limit=10&page=0&sort=created&order=DESC`

jvntf:
To clarify earlier comment, adding the 'see more' functionality would be a significant feature add, so may be best left alone for now. Added fetch by date created @LanesGood

LanesGood (approved):
@jvntf looks good!

PR #626, opened 2020-10-16
Refactor docker-compose to address problems with volume mounting

### LINKED ISSUE #625: postgresql /docker-entrypoint-initdb.d/* not reliable when volume mounted by docker-compose
In the docker-compose.yml, we are using volume mounts to control postgresql services like:

```yml
    volumes:
      - './.tmp/pg:/var/lib/postgresql/data'
      - ./scripts/create-test-db.sh:/docker-entrypoint-initdb.d/create-test-db.sh
```

and 
```yml
    volumes:
      # db will load this directory at start
      - ./scripts/osmesa-mock/init.sql:/docker-entrypoint-initdb.d/init.sql
```

These volume mounts do not work reliably, on @necoline computer the scripts in /docker-entrypoint-initdb. fail with "permission denied when running ...". This seems to be a somewhat common problem from discussions on stackoverflow etc.

Rather than troubleshoot [the volume mounting options, there are many](https://docs.docker.com/compose/compose-file/#volumes), we can try creating Dockerfiles that COPY the source scripts into place on the image, avoiding the volume mount issue entirely. Which is the traditional Docker method of doing this.

guidorice:
Closes #625 

Note: this PR also eliminates the local directory ` .tmp/pg/` and instead the scoreboard database is internal to the postgresql container.


necoline (approved):
Thank you for leading this out Alex!!

Issue #628, opened 2020-10-19
Only highlight user name in maproulette task progress table if user exists in app


LanesGood:
For new MapRoulette campaigns, if a user doesn't exist in the app database their name should not be a link.

As a user, I can click on a user's name in MapRoulette "challenge task progress" table to see their user page if they have one. If they don't have a user page, I am not able to click on their name.

![image](https://user-images.githubusercontent.com/12634024/96497557-deb74580-1218-11eb-9688-0b9d54954579.png)

LanesGood:
Closed by #631

PR #631, opened 2020-10-22
Fix/mr hashtags

jvntf:
Mix issue with maproulette challenge hashtags, typos, incorrect IDs.

### DIFF
--- a/components/common/Table.js
+++ b/components/common/Table.js
@@ -62,8 +62,9 @@ function selectCellFormatter (datatype, idMap, countryMap, campaignMap) {
       return (props) => {
         const { cell: { value } } = props
         const disabled = props.data[ props.row.index ].disableLink
+
         return disabled ? value : (
-          <Link ref={`/users/${idMap[value]}`}>
+          <Link href={`/users/${idMap[value]}`}>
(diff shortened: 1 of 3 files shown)


components/common/Table.js
```diff
@@ -174,7 +177,7 @@ export default function Table (props) {
                   {
                     row.cells.map(cell => {
                       return <td {...cell.getCellProps()} className={!isNaN(cell.value) && parseInt(cell.value) >= 0 ? 'table-align-right' : ''}>
-                        {cell.render('Cell')}
+                        {cell.render('Cell', 'test')}
```
LanesGood, line 180: @jvntf should this 'test' be here?

LanesGood (approved):
@jvntf this works well and the challenge pages are looking good. Can you confirm the following?
1. This PR fixes #628 
2. I see no osmesa stats tables in any of the campaigns. I'm assuming this is difficult to test. Are there any ways we can see the crossover between maproulette and osmesa in local dev?

jvntf:
@LanesGood yes this fixes #628. yes it is tough to test since we are using live data from maproulette, not sure of an easy way to test in local dev

PR #639, opened 2020-12-01
Format MapRoulette avgtimespent

LanesGood:
This PR converts AvgTimeSpent miliseconds to $m $s format in the Panel component and the table.

I'm looking into right-formatting the avgtimespent column, if possible, even though it is now a string.


LanesGood:
I've attempted to right-align the AvgTimeSpent column in the MapRoulette campaign page "Challenge Task Progress" table, but had no luck. Marking this as ready for review:

LanesGood:
Update - correctly formatted. Noting that the Table.js has become very difficult to read:

image

kamicut (approved):

Issue #440, opened 2019-11-05
Include active campaigns in country on country page


LanesGood:
The Country page does not currently include a large amount of useful information. The dashboard/user page redirects to the country page when a user has not yet mapped in countries, yet there are no actions that a user can take from this page. To improve the utility of the country page and to present information in a different and new way, the country page should include all active campaigns as cards in a new "Active Campaigns" section.

PR #646, opened 2021-02-03
Add alphabetical country sort to campaign filters

LanesGood:
I started stubbing out an alphabetical "Sort by Country" filter to the campaign filters based on #523. However, I realized that we're not actually getting the country information for a campaign from the tasking manager when ingesting the data.

It does appear that the country information is available per campaign (project) in the new tasking manager v2 API for TM4. Depending on the request URL, this is either returned as the `country` or `countryTag`. I'm unclear whether country is available in the v1 API for TM3 (and TM2?). I do see the country displayed on the project pages of TM3-created projects like this one: https://tasks.hotosm.org/projects/204, and it appears that the v2 API can be used to access those projects: https://hotosm.github.io/swagger/?url=https://tasking-manager-tm4-production-api.hotosm.org/api/v2/system/docs/json/#/projects/get_api_v2_projects__project_id__

I attempted to add the country to the API tasker service in `api/src/services/tm_types/tm3.js`, where I added `obj.country = country` in `updateDB`. This results in an error `error: column "country" of relation "campaigns" does not exist` when attempting to run `yarn seed`. [I'm not pushing this code due to the error]

To add this to the scoreboard database I'm assuming we need to add a migration, and then update the tasker files correctly in the API. Once this is complete, we'll actually be able to add the country to the campaign page, and complete this PR as well as fill in some other past ideas (#440). 

@kamicut or @jvntf would you be able to help get country data associated with the campaigns?

### DIFF
--- a/api/src/routes/campaigns.js
+++ b/api/src/routes/campaigns.js
@@ -77,6 +77,12 @@ module.exports = async (req, res) => {
       case 'Alphabetical Z-A':
         query = query.orderBy('name', db.raw('desc NULLS LAST'))
         break
+      case 'Countries A-Z':
+        query = query.orderBy('country', db.raw('asc NULLS LAST'))
+        break
+      case 'Countries Z-A':
+        query = query.orderBy('country', db.raw('desc NULLS LAST'))
+        break
     }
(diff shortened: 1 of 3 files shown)


LanesGood:
Per discussion, we're pausing development on this as we wait for TM4 integration.

PR #653, opened 2021-04-09
TM4 interface fixes

kamicut:
- Fixes prefixes for the TM4 interface
- Adds TM4 to select in admin page
- Fix tests
- Filters out draft campaigns from frontpage map


api/src/db/seeds/test/campaigns.js
```diff
@@ -13,7 +13,7 @@ exports.seed = async (knex) => {
     dirname: path.join(__dirname, '..', '..', '..', '..', 'tests', 'tapes')
   }))
 
-  const tm4proxy = http.createServer(yakbak('https://tasks-backend.openstreetmap.us/', {
+  const tm4proxy = http.createServer(yakbak('https://tasks-backend.openstreetmap.us/api', {
```
LanesGood, line 16: @kamicut I see a response for the original URL, but a 404 for this change.

kamicut: Yeah, it's only used in the construction of API urls as a prefix, for example https://tasks-backend.openstreetmap.us/api/v2/projects

LanesGood:
@kamicut in local development, I can run an instance of Scoreboard with 550 campaigns, and only one user. All the campaigns are MapRoulette. I've gone through the steps of yarn seed and migrate, but seed fails as usual for me and migrate doesn't throw any errors.

LanesGood (approved):
@kamicut thanks, let's merge it in! 🚀

PR #708, opened 2022-11-01
update timeseries data using submit button

kamicut:
This removes a little bit of the lag in the inputs. Do you see a problem with this approach @maxgrossman? I know you have this commit https://github.com/developmentseed/scoreboard/pull/702/commits/eb14fd4ed3c5ef19a11c6e71ffb778ba131c3165 that does the opposite but I didn't understand why you used a nextState variable.


maxgrossman (approved):
LGTM. 

yeah not sure rational for nextState either. if you want to just make it this.state go for it.

pages/timeseries.js
```diff
@@ -465,6 +468,7 @@ export default class TimeSeries extends Component {
                     className='basic-multi-select'
                     classNamePrefix='select' />
                 </fieldset>
+                <input className='button button--primary' type='button' value='Query' onClick={this.updateTimeseriesData} />
```
LanesGood, line 471:
```suggestion
                <input className='button button--primary' type='submit' value='Submit' onClick={this.updateTimeseriesData} />
```
@kamicut should this button type be a submit for the form? Regardless, I think "Submit" could be a better button value text here.

PR #709, opened 2022-11-04
Restrict timeseries chart to top 20 users

LanesGood:
This PR restricts the Time Series Edits Chart to the top users, sorted by edit, up to 20. The limit of 20 users is somewhat arbitrary, set only by the amount of vertical space available for the legend.

image

### DIFF
--- a/components/charts/TimeSeriesEditsChart.js
+++ b/components/charts/TimeSeriesEditsChart.js
@@ -10,6 +10,8 @@ const chartCrosswalk = {
   'km_coastlines_add_mod': 5
 }
 
+const MAX_USERS_CHART = 20
+
 // builds log scale inclusive of largest statistic
 function getLog10Scale (maxInteger) {
   const linearScale = []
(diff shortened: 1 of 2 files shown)


LanesGood:
Marking this as ready for review @maxgrossman @kamicut. I think the main question here is: do admins want to see timeseries charts with more than 20 users? If so, this PR can be closed. However, I think that the number of bars when there are more than 20 users becomes overwhelming.

kamicut (approved):
This looks good to me and ready to merge, but I think @maxgrossman should reply to the user case question. I suggest we merge and then work on an enhancement for the over 20 users use case (maybe paginate by using the `.slice` as a window into the users array @LanesGood).

maxgrossman:
I think this is cool. I think that 'top x' users makes lots of sense when filtering for edits by big teams or in a country, but less so when I am just looking for a few users. 

Maybe if the query returns less than 20 users 'top x users' gets omitted.
Think this could be addressed later. future paginating can address this.

PR #711, opened 2022-11-08
Make timeseries an admin area page

LanesGood:
This PR:
- Makes `/timeseries` accessible through the Admin panel (please feel free to suggest alternative text - currently "Run Timeseries Analysis")
- Removes link to timeseries from the standard Navbar
- Adds Admin label to timeseries page header

image

image


maxgrossman (approved):
awesome.

